PR #3, opened 2023-05-30
Added .gitignore

IfSentient:
Added missing .gitignore, fixed a couple of broken links in the README.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,27 @@
+i# Binaries for programs and plugins
+*.exe
+*.exe~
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -0,0 +1,27 @@
+i# Binaries for programs and plugins
```
ferruvich, line 1: Just noticed this `i` - is it intended?

IfSentient: Nope, vim artifact 🤦

PR #4, opened 2023-05-31
Export TS types in cuetsy config for the TS Jenny.

IfSentient:
Export TS types in cuetsy config for the TS Jenny.


joeblubaugh:
Thank you for debugging and fixing :-)

PR #6, opened 2023-05-31
Add Initial GitHub Actions for PR, Main, and Release

IfSentient:
These are initial, tested workflows for pull requests (lint, test->build), merges to main (test->build), and release (test->build->release).

They have been tested on a fork and work correctly.

Also added in this PR are some dot config files that were missing from the initial commit (copied from archived version). 

Drone information in the Makefile has been removed.


.dependabot.yml
radiohead, line 1: This might have been misconfigured previously – this needs to live in `.github/dependabot.yml` instead.

radiohead (commented):
Looks good, just a small issue with Dependabot config.

.dependabot.yml
IfSentient: Ah, thanks, good catch. Moved the file.

radiohead (approved):

Issue #14, opened 2023-06-02
Kubernetes Patch Operations are not translated


IfSentient:
The kubernetes client (`k8s.Client`,`k8s.SchemalessClient`) translates to/from the `resource.Object` representation of metadata from/to the kubernetes metadata. However, patch operations are passed along as-is. If a user uses a path which is correct in the representation of the metadata (say, `/metadata/createdBy`), but is represented differently in kubernetes metadata, that patch operation will not do what they expect (and may fail).

To fix this, we should translate the paths in a `resource.PatchRequest` _if and only if_ they reference fields which are not present in the kubernetes metadata. So, for example:
`/metadata/createdBy` -> `/metadata/annotations/grafana.com~1createdBy`
`/metadata/finalizers` -> `/metadata/finalizers` (no translation)
We _also_ should translate any keys in `extraFields` into a corresponding kubernetes field:
`/metadata/extraFields/managedFields/manager` -> `/metadata/managedFields/manager`

This way, if the patch operation already uses the kubernetes paths, nothing will be changed, but if they use paths based on `resource.CommonMetadata` and the custom metadata, we will translate into the kubernetes representation of that data.

PR #15, opened 2023-06-02
[bugfix] Translate patch operations for kubernetes

IfSentient:
Translate patch requests in translation layer, as metadata fields that are custom, non-kubernetes common, or are in extraFields need to have their paths updated if the provided path is based on the resource's metadata object. If the provided path uses the kubernetes path, the translation layer will not change it.

This came up while working on admission control, but I think it warrants a separate PR before that upcoming one, which is still in-progress.

Resolves https://github.com/grafana/grafana-app-sdk/issues/14

### DIFF
--- a/k8s/translation.go
+++ b/k8s/translation.go
@@ -231,8 +231,8 @@ func getV1ObjectMetaFields() map[string]struct{} {
 		if len(jsonTag) == 0 || jsonTag[0] == '-' || jsonTag[0] == ',' {
 			continue
 		}
-		if strings.Index(jsonTag, ",") > 0 {
-			jsonTag = jsonTag[:strings.Index(jsonTag, ",")]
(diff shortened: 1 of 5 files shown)


k8s/translation.go
```diff
@@ -179,6 +179,66 @@ func marshalJSON(obj resource.Object, extraLabels map[string]string, cfg ClientC
 	return json.Marshal(co)
 }
 
+var metaV1Fields = getV1ObjectMetaFields()
```
IfSentient, line 182: Done dynamically once with reflection, instead of hard-coding the v1meta.ObjectMeta fields (can't use an empty JSON-marshaled object for keys, because they nearly all have `,omitempty` on their JSON tags).

radiohead (approved):

PR #18, opened 2023-06-07
[tutorial] Update Shell Commands to allow for GitHub Copy/Paste

IfSentient:
Remove leading '$' from standalone shell commands to enable copy-paste from GH UI. Reformat some command blocks to also allow for this.


ferruvich (approved):
LGTM

PR #23, opened 2023-06-09
[SimpleStore] Ensure Update call doesn't erase existing metadata

IfSentient:
In resource.SimpleStore.Update, make a Get call first to get the current metadata (as only the spec is passed), then run the ObjectMetadataOptions on the resulting metadata. Otherwise, the subsequent PUT call in the underlying Client erases the existing non-managed metadata (such as finalizers).

### DIFF
--- a/resource/simplestore.go
+++ b/resource/simplestore.go
@@ -131,6 +131,21 @@ func (s *SimpleStore[T]) Update(ctx context.Context, identifier Identifier, obj 
 	object := SimpleObject[T]{
 		Spec: obj,
 	}
+	// Before we can run the opts on the metadata, we need the current metadata
+	// TODO: should this whole thing instead be serialized to a patch?
+	// It could affect expected behavior, though, as WithResourceVersion makes sure it matches the RV you supply
+	current, err := s.Get(ctx, identifier)
+	if err != nil {
+		return nil, err
+	}
+	object.CommonMeta = current.CommonMetadata
+	customFields := current.CustomMetadata.MapFields()
+	if len(customFields) > 0 {
+		object.CustomMeta = make(SimpleCustomMetadata)
+		for f, v := range customFields {
+			object.CustomMeta[f] = v
+		}
+	}
 	for _, opt := range opts {
 		opt(&object.CommonMeta)
 	}
--- a/resource/simplestore_test.go
+++ b/resource/simplestore_test.go
@@ -170,8 +170,21 @@ func TestSimpleStore_Update(t *testing.T) {
 		Name:      "test",
 	}
 
-	t.Run("error", func(t *testing.T) {
+	t.Run("get error", func(t *testing.T) {
 		cerr := fmt.Errorf("I AM ERROR")
+		client.GetFunc = func(ctx context.Context, identifier Identifier) (Object, error) {
+			return nil, cerr
+		}
+		ret, err := store.Update(ctx, id, "")
+		assert.Nil(t, ret)
+		assert.Equal(t, cerr, err)
+	})
+
+	t.Run("update error", func(t *testing.T) {
+		cerr := fmt.Errorf("SOY ERROR")
+		client.GetFunc = func(ctx context.Context, identifier Identifier) (Object, error) {
+			return retObj, nil
+		}
 		client.UpdateFunc = func(ctx context.Context, identifier Identifier, obj Object, options UpdateOptions) (Object, error) {
 			return nil, cerr
 		}
@@ -181,6 +194,9 @@ func TestSimpleStore_Update(t *testing.T) {
 	})
 
 	t.Run("success, no metadata options", func(t *testing.T) {
+		client.GetFunc = func(ctx context.Context, identifier Identifier) (Object, error) {
+			return retObj, nil
+		}
 		client.UpdateFunc = func(c context.Context, identifier Identifier, obj Object, options UpdateOptions) (Object, error) {
 			assert.Equal(t, ctx, c)
 			assert.Equal(t, id, identifier)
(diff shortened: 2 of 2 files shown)


joeblubaugh:
@cesararmada does this affect your work at all?

cesararmada:
> @cesararmada does this affect your work at all?

Yes, this fixes the bug I was trying to solve.

radiohead (commented):
@IfSentient I'm not sure this is the correct approach here. I agree with TODO however – this could work if we'd switch to patching instead of updating.

IMHO we should either:

a) Keep current semantics of `Update` intact and add a new `Patch` (?) method with patch semantics.
b) Rename `Update` to `Patch` (?) and change it to use patch semantics.
C) Keep current semantics of `Update` intact and add a new `Merge` (?) method. `Merge` would work with Get-and-Put semantics.

i.e. the way I see it `Update == PUT`, so the caller must keep track of the client vs. server state themselves but in return they get better performance (i.e. calling `Update` results in a single API call).

In general I would prefer not to rely on get-and-put semantics unless absolutely necessary, because they create a lot of server-side overhead.

IfSentient:
> @IfSentient I'm not sure this is the correct approach here. I agree with TODO however – this could work if we'd switch to patching instead of updating.
> 
> IMHO we should either:
> 
> a) Keep current semantics of `Update` intact and add a new `Patch` (?) method with patch semantics. b) Rename `Update` to `Patch` (?) and change it to use patch semantics. C) Keep current semantics of `Update` intact and add a new `Merge` (?) method. `Merge` would work with Get-and-Put semantics.
> 
> i.e. the way I see it `Update == PUT`, so the caller must keep track of the client vs. server state themselves but in return they get better performance (i.e. calling `Update` results in a single API call).
> 
> In general I would prefer not to rely on get-and-put semantics unless absolutely necessary, because they create a lot of server-side overhead.

@radiohead the big problem with the PUT semantics of Update in SimpleStore is that SImpleStore is designed to only consume the spec object (in order to be "simple"), so the only way to pass along metadata is via get-and-put. This isn't a problem for TypedStore and Store, because they take whole objects in their Update methods.

Also note that get-and-put is done _anyway_ if a ResourceVersion is not present in the UpdateOptions for _any_ Client.Update interaction (via stores or directly), because for CRD's in kubernetes, leaving the RV empty to implicitly use latest doesn't work.

radiohead:
> > @IfSentient I'm not sure this is the correct approach here. I agree with TODO however – this could work if we'd switch to patching instead of updating.
> > IMHO we should either:
> > a) Keep current semantics of `Update` intact and add a new `Patch` (?) method with patch semantics. b) Rename `Update` to `Patch` (?) and change it to use patch semantics. C) Keep current semantics of `Update` intact and add a new `Merge` (?) method. `Merge` would work with Get-and-Put semantics.
> > i.e. the way I see it `Update == PUT`, so the caller must keep track of the client vs. server state themselves but in return they get better performance (i.e. calling `Update` results in a single API call).
> > In general I would prefer not to rely on get-and-put semantics unless absolutely necessary, because they create a lot of server-side overhead.
> 
> @radiohead the big problem with the PUT semantics of Update in SimpleStore is that SImpleStore is designed to only consume the spec object (in order to be "simple"), so the only way to pass along metadata is via get-and-put. This isn't a problem for TypedStore and Store, because they take whole objects in their Update methods.

Fair point – in such case I suggest that we switch to using PATCH instead of PUT.

> Also note that get-and-put is done _anyway_ if a ResourceVersion is not present in the UpdateOptions for _any_ Client.Update interaction (via stores or directly), because for CRD's in kubernetes, leaving the RV empty to implicitly use latest doesn't work.

Yeah and I'm not a big fan of this, but it's a separate topic for discussion.

Let me expand on my reasoning – I tend to think of Kubernetes API as just another storage layer with optional write serialisability. If I were using e.g. SQL I can choose whether I want to force-write some data for efficiency's sake vs. using a transaction for read-compare-update semantics.

Now, Kubernetes API is unfortunately opinionated in that it doesn't provide its API clients with the ability to forcefully apply a specific configuration via the PUT / update verb, i.e. one has to always supply the correct RV. However, in order to emulate the forced write we could use the PATCH verb instead and that'd be my preferred approach here.

IfSentient:
> > > @IfSentient I'm not sure this is the correct approach here. I agree with TODO however – this could work if we'd switch to patching instead of updating.
> > > IMHO we should either:
> > > a) Keep current semantics of `Update` intact and add a new `Patch` (?) method with patch semantics. b) Rename `Update` to `Patch` (?) and change it to use patch semantics. C) Keep current semantics of `Update` intact and add a new `Merge` (?) method. `Merge` would work with Get-and-Put semantics.
> > > i.e. the way I see it `Update == PUT`, so the caller must keep track of the client vs. server state themselves but in return they get better performance (i.e. calling `Update` results in a single API call).
> > > In general I would prefer not to rely on get-and-put semantics unless absolutely necessary, because they create a lot of server-side overhead.
> > 
> > 
> > @radiohead the big problem with the PUT semantics of Update in SimpleStore is that SImpleStore is designed to only consume the spec object (in order to be "simple"), so the only way to pass along metadata is via get-and-put. This isn't a problem for TypedStore and Store, because they take whole objects in their Update methods.
> 
> Fair point – in such case I suggest that we switch to using PATCH instead of PUT.

The only problem I see here (and why I didn't use PATCH originally in the `Update` method) is that we can't do optimistic resource locking with ResourceVersion. I'm OK with saying that `SimpleStore` _doesn't_ do resource locking, and if you want to use that, you have to use `TypedStore` or `Store`, but I just want to make sure this is clear for all folks.

@cesararmada @joeblubaugh IIRC the `resource.WithResourceVersion` option was added for SLO app purposes, is it ok if we make this change to disallow ResourceVersion locking in the `SimpleStore` and refer folks to using `TypedStore` or `Store` for it? `TypedStore` is actually very easy to use with an `operator.Watcher`, as the type you supply it is the same type that comes in from the watcher events.

radiohead (approved):
Approving this to unblock the bugfix.

PR #29, opened 2023-06-15
Admission Control

IfSentient:
This PR introduces methods of doing validating and mutating admission control via kubernetes webhooks, expressed in a generic manner. It focuses on four parts:
1. A generic set of interfaces and structs for admission control (in the `resource` package)
2. A kubernetes webhook server which can run as a controller, and consumes handlers for mutating/validating kinds
3. Opinionated mutating/validating controllers which handle essential grafana kind metadata for you
4. `project local generate` updates to allow for the generated manifests to include webhooks. This includes some changes to the Tiltfile to make sure we don't run into weird blocking scenarios for startup.

(2) is the largest component of this, and most of the complexity is around (a) translating kubernetes, and (b) associating specific mutating/validating controllers with kinds. Translating kubernetes is of particular complexity, because we need to _also_ handle limitations in JSONPatch w.r.t. kubernetes when it comes to annotations--in essence, we need to modify the patch operations returned to us by the controllers based on whether or not it has non-kubernetes metadata which must be encoded in annotations, and if annotations is currently absent/an empty map. This is a large PR, though more than half of the lines of code added should be from tests.

I'm still not 100% happy with the implementation in the `k8s` package, but I think the exposed interfaces/types are fine going forward, and we may just need to tinker with the logic in the functions themselves, so it would be transparent to users.


ferruvich (commented):
> Do Admission Handlers need to exist, or am I creating an extra layer of abstraction? Their initial reason was to separate out the controller association and marshal/unmarshal logic from actual HTTP, so that other means of calling the webhooks, such as over an RPC connection, could easily be tacked on, and a different operator.Controller could exist for different protocol mechanisms, while still using the same underlying controller association and marshal/unmarshal logic. But it does add complexity to the whole thing. The alternative is to merge the handlers with the Webhook Server itself, and do all this in one place.

AFAICS, here we're talking about the `k8s` implementation, am I right? As for the interfaces I can see a consumer just implementing `resource.(Validating|Mutating)AdmissionController` and use them directly in the `operator.Operator` implementation.

That said, do we see any use case where operators using _different_ protocols use the same identical `resource.(Validating|Mutating)AdmissionController` implementation? In such case I can see the usefulness of Handlers. I personally see this kinda of an edge case, where we need to expose the same identical admissionControllers to different sources, using different protocols. 
wdyt?

> How controllers get added to the server. Do we:
> 1 Have them supplied at creation time, as part of the config (this approach can be seen in the PR)? This feels a bit clunky to use.
> 2 Have the server expose Add functions for registering controllers, and the ability to set the default controllers directly (this is partially present in the PR)? This is a bit clunky because the server is just wrapping the handlers. However, if we dispose of separate handlers, then this becomes much simpler.
> 3 Have the server expose the underlying ValidatingAdmissionHandler and MutatingAdmissionHandler so the user can call methods on them directly and/or expose methods for setting the handlers themsevles (rather than exposing them)? Starting on this approach really made me feel like perhaps the handlers are "too much abstraction"

I believe I've written my opinion in a comment on this PR, but I'll reiterate it here:
My preferred approach would be to have a combination of both 1 and 2 where controllers are added in `Config` at creation time + having the possibility to register new controllers through functions.

k8s/webhooks.go
```diff
@@ -0,0 +1,124 @@
...
+	// in their Opinionated variants. This is done at setup time.
+	WrapControllers bool
+}
+
+type TLSConfig struct {
+	CertPath string
+	KeyPath  string
+}
+
+func NewWebhookServer(config WebhookServerConfig) (*WebhookServerController, error) {
```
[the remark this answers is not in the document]
IfSentient: Yeah, the naming is a bit confusing because it was as I was waffling between the two possible new functions, and I had re-named the struct from `WebhookServer` to `WebhookServerController`. I'll make sure whatever new function seems best is named `NewWebhookServerController`.

I think your suggestion makes sense, thanks!

IfSentient:
> AFAICS, here we're talking about the k8s implementation, am I right? As for the interfaces I can see a consumer just implementing resource.(Validating|Mutating)AdmissionController and use them directly in the operator.Operator implementation.

Yeah, I should have been more clear. I think the stuff in `resource` is pretty well-defined, but I was waffling on the extra layers I wrote in `k8s` itself. `k8s.*Handler` structs aren't used directly, so I felt like they may have been unnecessary abstraction.

> That said, do we see any use case where operators using different protocols use the same identical resource.(Validating|Mutating)AdmissionController implementation? In such case I can see the usefulness of Handlers. I personally see this kinda of an edge case, where we need to expose the same identical admissionControllers to different sources, using different protocols.
wdyt?

It's definitely an edge case for now, but I was thinking about a future where admission control could be inlined in kubernetes-adjacent scenarios where the operator is running as a WASM process, or a managed binary with an RPC communication channel from the API server, so there would be more efficient manners of communication than using HTTPS. Might be premature optimization which can be changed later, since it would be hidden behind implementation.

k8s/admission.go
```diff
@@ -0,0 +1,232 @@
...
+			Value:     op.Value,
+		}
+	}
+	return &resp, nil
+}
+
+// OpinionatedMutatingAdmissionController is a MutatingAdmissionController which wraps an optional user-defined
+// Mutate() function with a set of additional PatchOperations which set metadata and label properties.
+type OpinionatedMutatingAdmissionController struct {
+	MutateFunc func(request *resource.AdmissionRequest) (*resource.MutatingResponse, error)
```
radiohead, line 78: How about we wrap the `MutatingAdmissionController` interface instead?

k8s/admission.go
```diff
@@ -0,0 +1,232 @@
...
+	return &OpinionatedMutatingAdmissionController{
+		MutateFunc: mutateFunc,
+	}
+}
+
+// OpinionatedValidatingAdmissionController implements resource.ValidatingAdmissionController and performs initial
+// validation on reserved metadata fields which are stores as annotations in kubernetes, ensuring that if any changes are made,
+// they are allowed, before calling the underlying admission validate function.
+type OpinionatedValidatingAdmissionController struct {
+	ValidateFunc func(*resource.AdmissionRequest) error
```
radiohead, line 180: Same here – IMHO wrapping the interface makes more sense, that way we can have consistent patter of nesting the controllers.

k8s/webhooks.go
```diff
@@ -0,0 +1,377 @@
...
+// It implements operator.Controller and can be run as a controller in an operator, or as a standalone process.
+type WebhookServer struct {
+	// DefaultValidatingController is the default ValidatingAdmissionController to use if one is not defined for the schema in the request.
+	// If this is empty, the request will be rejected.
+	DefaultValidatingController resource.ValidatingAdmissionController
+	// DefaultMutatingController is the default MutatingAdmissionController to use if one is not defined for the schema in the request.
+	// If this is empty, the request will be rejected.
+	DefaultMutatingController resource.MutatingAdmissionController
+	validatingControllers     map[string]validatingAdmissionControllerTuple
+	mutatingControllers       map[string]mutatingAdmissionControllerTuple
```
radiohead, line 54: I remember we've discussed that having multiple controllers per Kind doesn't make sense, but if we introduce system-level ones (e.g. the logic in Opinionated controllers) then it might in fact make sense to support more than 1 controller per Kind and execute them concurrently (at least for validation).

IfSentient: Yeah, since you can't really execute mutates in parallel, I don't think having multiple mutating controllers should be allowed (if you want that, just chain controllers yourself so you can guarantee ordering). I'll make some updates to allow multiple validating controllers which can be executed in parallel on validate calls.

IfSentient: On further reflection, this gets a bit complicated, namely for error handling reasons. If we have multiple errors from different validators, how do we combine them? Each error can have its own required status code, and conflicting status codes gets weird. I think chaining, like with middleware, makes more sense here, at least for the immediate term.

radiohead: I'm fine with chaining them for now, but for the future I think we could either bail on the first error (which luckily `errgroup` already does) or set up some sort of error hierarchy. I'd personally prefer the former.

sdboyer:
Do we want to try to unite this with grafana/kindsys#22, or let these happen independently for now and treat converging them as an (immediate, i hope) follow up?

k8s/admission.go
```diff
@@ -84,104 +54,65 @@ var now = time.Now
...
+		resp, err = o.Underlying.Mutate(ctx, request)
 		if err != nil {
 			return resp, err
 		}
 	}
-	if resp == nil || resp.PatchOperations == nil {
+	if resp == nil || resp.UpdatedObject == nil {
 		resp = &resource.MutatingResponse{
-			PatchOperations: make([]resource.PatchOperation, 0),
+			UpdatedObject: request.Object,
```
radiohead, line 69: Not sure if this is an issue but it's possible that we want to copy the object here instead of using a reference to it, so that we don't mutate the request object – I don't know if anything else depends on it.

k8s/webhooks.go
```diff
@@ -197,7 +197,7 @@ func (w *WebhookServer) HandleValidateHTTP(writer http.ResponseWriter, req *http
 	}
 
 	// Run the controller
-	err = controller.Validate(admReq)
+	err = controller.Validate(context.Background(), admReq)
```
radiohead, line 200: We should be passing the real context here:
```suggestion
	err = controller.Validate(req.Context(), admReq)
```

k8s/webhooks.go
```diff
@@ -239,6 +239,7 @@ func (w *WebhookServer) HandleMutateHTTP(writer http.ResponseWriter, req *http.R
 	// Unmarshal the admission review
 	admRev, err := unmarshalKubernetesAdmissionReview(body, resource.WireFormatJSON)
 	if err != nil {
+		fmt.Println(err)
```
radiohead, line 242: Leftovers?

k8s/webhooks.go
```diff
@@ -271,49 +272,16 @@ func (w *WebhookServer) HandleMutateHTTP(writer http.ResponseWriter, req *http.R
 	}
 
 	// Run the controller
-	mResp, err := controller.Mutate(admReq)
+	mResp, err := controller.Mutate(context.Background(), admReq)
```
radiohead, line 275: Same here – we should use the real context:
```suggestion
	mResp, err := controller.Mutate(req.Context(), admReq)
```

k8s/webhooks_test.go
```diff
@@ -232,45 +252,42 @@ func TestWebhookServer_HandleMutateHTTP(t *testing.T) {
...
+			payload:   admissionRequestBytesNoDefaults,
 			// Patch is base64-encoded
-			// [{"path":"/foo/bar","op":"replace","value":"foobar"}] => W3sicGF0aCI6Ii9mb28vYmFyIiwib3AiOiJyZXBsYWNlIiwidmFsdWUiOiJmb29iYXIifV0=
-			expectedResponse:   []byte(`{"response":{"uid":"foo","allowed":true,"patchType":"JSONPatch","patch":"W3sicGF0aCI6Ii9mb28vYmFyIiwib3AiOiJyZXBsYWNlIiwidmFsdWUiOiJmb29iYXIifV0="}}`),
+			// [{"op":"add","path":"/metadata/annotations","value":{"grafana.com/createdBy":"me","grafana.com/updateTimestamp":"2023-07-06T16:49:10-04:00","grafana.com/updatedBy":"you"}}] =>
+			// W3sib3AiOiJhZGQiLCJwYXRoIjoiL21ldGFkYXRhL2Fubm90YXRpb25zIiwidmFsdWUiOnsiZ3JhZmFuYS5jb20vY3JlYXRlZEJ5IjoibWUiLCJncmFmYW5hLmNvbS91cGRhdGVUaW1lc3RhbXAiOiIyMDIzLTA3LTA2VDE2OjQ5OjEwLTA0OjAwIiwiZ3JhZmFuYS5jb20vdXBkYXRlZEJ5IjoieW91In19XQ==
+			expectedResponse:   []byte(`{"response":{"uid":"foo","allowed":true,"patchType":"JSONPatch","patch":"W3sib3AiOiJhZGQiLCJwYXRoIjoiL21ldGFkYXRhL2Fubm90YXRpb25zIiwidmFsdWUiOnsiZ3JhZmFuYS5jb20vY3JlYXRlZEJ5IjoibWUiLCJncmFmYW5hLmNvbS91cGRhdGVUaW1lc3RhbXAiOiIyMDIzLTA3LTA2VDE2OjQ5OjEwLTA0OjAwIiwiZ3JhZmFuYS5jb20vdXBkYXRlZEJ5IjoieW91In19XQ=="}}`),
```
radiohead: Looks like this is failing.

k8s/webhooks.go
IfSentient: Oops!

k8s/webhooks_test.go
IfSentient: Isn't failing locally 🤔

IfSentient: Time zones 🤦

k8s/admission.go
```diff
@@ -0,0 +1,166 @@
...
+		cmd.UpdateTimestamp = now()
+	default:
+		// Do nothing
+	}
+	resp.UpdatedObject.SetCommonMetadata(cmd)
+	return resp, nil
+}
+
+// NewOpinionatedMutatingAdmissionController creates a pointer to a new OpinionatedMutatingAdmissionController wrapping the
+// provided mutateFunc (nil mutateFunc argument is allowed, and will cause the controller to not call the underlying function)
```
charandas, line 97: Looks like the parameter here is no longer `mutateFunc`. I understand that its coming from `wrap.MutateFunc`? The comment reads like it was written for an older form of this function.

k8s/admission.go
radiohead: Fair enough 👍

radiohead (approved):

k8s/admission.go
IfSentient: Ah, good catch, let me update the docs

radiohead (approved):

PR #38, opened 2023-06-27
Don't attempt to build the plugin if files are missing

### LINKED ISSUE #37: When not building a plugin, `make` should not fail
When code generating for a resource/operator only, i.e. no plugin code, running `make` should not fail.

bcotton:
Check for the existance of plugin.json or Magefile.go before attempting to build the plugin. Fixes #37

### DIFF
--- a/cmd/grafana-app-sdk/templates/Makefile.tmpl
+++ b/cmd/grafana-app-sdk/templates/Makefile.tmpl
@@ -36,12 +36,16 @@ build/plugin: build/plugin-backend build/plugin-frontend
 build/plugin-frontend:
 ifeq ("$(wildcard plugin/plugin.json)","plugin/plugin.json")
 	@cd plugin && yarn install && yarn build
+else
+	@echo "No plugin.json found, skipping frontend build"
 endif
 
 .PHONY: build/plugin-backend
 build/plugin-backend:
 ifeq ("$(wildcard plugin/Magefile.go)","plugin/Magefile.go")
 	@cd plugin && mage -v
+else
+	@echo "No Magefile.go found, skipping backend build"
 endif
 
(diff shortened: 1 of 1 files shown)


cmd/grafana-app-sdk/templates/Makefile.tmpl
```diff
@@ -34,11 +34,15 @@ build/plugin: build/plugin-backend build/plugin-frontend
 
 .PHONY: build/plugin-frontend
 build/plugin-frontend:
+ifeq ("$(wildcard plugin/plugin.json)","plugin/plugin.json")
 	@cd plugin && yarn install && yarn build
+endif
```
IfSentient, line 39: Could we add a message when we don't build this as well?
```suggestion
else
	@echo "Not building plugin frontend as plugin/plugin.json is missing"
endif
```

cmd/grafana-app-sdk/templates/Makefile.tmpl
```diff
@@ -34,11 +34,15 @@ build/plugin: build/plugin-backend build/plugin-frontend
...
 
 .PHONY: build/plugin-backend
 build/plugin-backend:
+ifeq ("$(wildcard plugin/Magefile.go)","plugin/Magefile.go")
 	@cd plugin && mage -v
+endif
```
IfSentient, line 45: Same idea as with frontend, just a message to let the user know the build target was skipped and why.
```suggestion
else
	@echo "Not building plugin backend as plugin/Magefile.go is missing"
endif
```

IfSentient (approved):
Looks good, just a couple of suggestions to add an echoed statement when the make targets aren't built because the plugin info is missing.

IfSentient (approved):

PR #40, opened 2023-06-27
Add NewSchemalessClient

toddtreece:
This adds a `NewSchemalessClient` function to allow the use of `SchemalessClient` outside of the `k8s` package.

### DIFF
--- a/k8s/schemaless.go
+++ b/k8s/schemaless.go
@@ -30,6 +30,8 @@ type SchemalessClient struct {
 
 // NewSchemalessClient creates a new SchemalessClient using the provided rest.Config and ClientConfig.
 func NewSchemalessClient(kubeConfig rest.Config, clientConfig ClientConfig) *SchemalessClient {
+	kubeConfig.NegotiatedSerializer = &GenericNegotiatedSerializer{}
+	kubeConfig.UserAgent = rest.DefaultKubernetesUserAgent()
 	return &SchemalessClient{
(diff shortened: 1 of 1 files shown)


k8s/schemaless.go
```diff
@@ -28,6 +28,15 @@ type SchemalessClient struct {
 	mutex sync.Mutex
 }
 
+// NewSchemalessClient creates a new SchemalessClient using the provided rest.Config and ClientConfig.
+func NewSchemalessClient(kubeConfig rest.Config, clientConfig ClientConfig) *SchemalessClient {
```
IfSentient, line 32: We should make sure we're using the correct NegotiatedSerializer for the underlying REST clients we generate, otherwise we can get bad/undefined behavior. We do the [same thing in NewClientRegistry](https://github.com/grafana/grafana-app-sdk/blob/a6198fc81f66622f0de678d489d4ada785e56540/k8s/client_registry.go#L14-L15).
```suggestion
func NewSchemalessClient(kubeConfig rest.Config, clientConfig ClientConfig) *SchemalessClient {
	kubeConfig.NegotiatedSerializer = &GenericNegotiatedSerializer{}
	kubeConfig.UserAgent = rest.DefaultKubernetesUserAgent()
```

IfSentient (changes requested):
Just one small change to make sure the client-go kubernetes client doesn't have trouble.

IfSentient (approved):
LGTM

PR #45, opened 2023-06-30
[Actions] Fix Linter Version

IfSentient:
Use the same version of golangci-lint in the GH actions as we do in the Makefile.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,7 +17,7 @@ jobs:
     - name: golangci-lint
       uses: golangci/golangci-lint-action@v3
       with:
-        version: latest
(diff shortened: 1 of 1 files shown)


radiohead (approved):

.github/workflows/pr.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: golangci-lint
       uses: golangci/golangci-lint-action@v3
       with:
-        version: latest
+        version: '1.51.2'
```
radiohead, line 20: Looks like the action requires `v` in the version string:
```suggestion
        version: v1.51.2
```

radiohead (approved):

Issue #1, opened 2023-05-30
Support Reconciler Pattern


IfSentient:
### What

The SDK should support the `ReconcilerController` alongside with the `InformerController`. The reconciler-based controller should loosely follow the Reconciler interface `github.com/kubernetes-sigs/controller-runtime`, i.e. something along the lines of:

```go
// Describes the action that triggered reconciliation.
type ReconcileAction int

const (
  ReconcileActionUnknown ReconcileAction = iota

  // Object was created
  ReconcileActionCreated

  // Object was updated
  ReconcileActionUpdated

  // Object was deleted
  ReconcileActionDeleted

  // Object was re-synced (i.e. periodic reconciliation)
  ReconcileActionResynced
)

// The request encapsulates both the action and the object state at the time of the action.
type ReconcileRequest struct {
  Action ReconcileAction
  Object resource.Object
}

// The result can be returned by the reconciler
// e.g. when reconciler wants to requeue the reconciliation
// because some remote component is not ready yet.
//
// By default it will be empty.
type ReconcileResult struct {
  RequeueAfter time.Duration
}

// If the reconciler returns an error, the request will be requeued.
// If the reconciler returns a non-empty result, the request will be requeued as well (according to RequeueAfter).
type Reconciler interface {
  Reconcile(context.Context, ReconcileRequest) (ReconcileResult, error)
}
```

### Why

It is a much simpler pattern, since most of the reconcilers don't care about the actions (and can thus safely ignore the action part of the request) since the logic is the same.

It also requires less boilerplate code, which is good.

PR #55, opened 2023-07-14
[operator] Introduce a Reconciler Pattern

IfSentient:
Introduce a Reconciler pattern for the InformerController, with an interface defining a Reconciler, and methods in the InformerController for adding Reconciler(s) for a given kind.

The Reconciler pattern allows for explicit failure (returning an error), which uses the normal RetryPolicy of the InformerController, *or* supplying a Retry-After response explicitly telling the InformerController to try this exact same Reconcile action again after the request interval has passed.

When using the Retry-After or explicit error, a State can also be provided in the Response which will be supplied as part of the retried Reconcile action. This allows a Reconcile to "partly complete" if there is state information which cannot be stored in the object itself (the most crucial example of this is if the Reconcile action performs an action with a third-party service, and then is unable to update the status in the API server). 

This PR also introduces an `OpinionatedReconciler`, which, much like the `OpinionatedWatcher`, uses automated finalizer logic to ensure events are not missed during downtime, and to re-map `Create`/`Add` events which are receieved on startup to `Sync` actions if the finalizer is already present.

It also introduces a _lot_ of lines of testing, apologies on the large PR size, the bulk of it is tests, and some of it is re-arrangement of the InformerController add/update/delete handler functionality to make it more readable and in smaller functions.

Resolves https://github.com/grafana/grafana-app-sdk/issues/1


operator/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+type ReconcileRequest struct {
+	// Action is the action that triggered this ReconcileRequest
+	Action ReconcileAction
+	// Object is the object art the time of the received action
+	Object resource.Object
+	// State is a user-defined map of state values that can be provided on retried ReconcileRequests.
+	// See State in ReconcileResult. It will always be nil on an initial Reconcile call,
+	// and will only be non-nil if a prior Reconcile call with this ReconcileRequest returned a State
+	// in its ReconcileResult alongside either a RequeueAfter or an error.
+	State map[string]any
```
radiohead, line 58: Not sure if it's worth adding this for extra complexity. Normally it's expected that the reconciler will handle storing intermediate state itself, by e.g. updating the object's status in the API.

operator/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+	client        PatchClient
+}
+
+const opinionatedReconcilerPatchStateKey = "grafana-app-sdk-opinionated-reconciler-create-patch-status"
+
+// Reconcile consumes a ReconcileRequest and passes it off to the underlying ReconcileFunc, using the following criteria to modify or drop the request if needed:
+//   - If the action is a Create, and the OpinionatedReconciler's finalizer is in the finalizer list, update the action to a Resync
+//   - If the action is a Create, and the OpinionatedReconciler's finalizer is missing, add the finalizer after the delegated Reconcile request returns successfully
+//   - If the action is an Update, and the DeletionTimestamp is non-nil, remove the OpinionatedReconciler's finalizer, and do not delegate (the subsequent Delete will be delegated)
+//   - If the action is an Update, and the OpinionatedReconciler's finalizer is missing (and DeletionTimestamp is nil), add the finalizer, and do not delegate (the subsequent update action will delegate)
```
radiohead, line 144: IMHO this logic should be in the watcher / informer, not in the reconciler.

operator/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+		finalizer: finalizer,
+		client:    client,
+	}, nil
+}
+
+// OpinionatedReconciler wraps an ordinary Reconciler with finalizer-based logic to convert "Created" events into
+// "resync" events on start-up when the reconciler has handled the "created" event on a previous run,
+// and ensures that "delete" events are not missed during reconciler down-time by using the finalizer.
+type OpinionatedReconciler struct {
+	ReconcileFunc func(context.Context, ReconcileRequest) (ReconcileResult, error)
```
radiohead, line 133: Same as in admission control – I suggest wrapping the `Reconciler` interface, rather than adding an extra function type.

operator/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+	// in its ReconcileResult alongside either a RequeueAfter or an error.
+	State map[string]any
+}
+
+// TypedReconciler is a variant of SimpleReconciler in which a user can specify the underlying type of the resource.Object
+// which is in the provided ReconcileRequest. Reconcile() will then attempt to cast the resource.Object in the
+// ReconcileRequest into the provided T type and produce a TypedReconcileRequest, which will be passed to ReconcileFunc.
+type TypedReconciler[T resource.Object] struct {
+	// ReconcileFunc is called by TypedReconciler.Reconcile using the T-typed Object instead of a resource.Object.
+	ReconcileFunc func(context.Context, TypedReconcileRequest[T]) (ReconcileResult, error)
```
radiohead, line 252: Again – I think we should wrap the interface, not a function.

radiohead (approved):
Looks good! I have a few opinionated comments but since I'm going to be out for the next few weeks and I don't see anything inherently wrong / broken I'm OK with merging this as-is.

operator/reconciler.go
IfSentient: I mostly agree, the `OpinionatedReconciler` is partly a stopgap until we can improve the Informer interface and implementation to understand resync events--though we'll still have the problem of "startup" events always being considered one event type (Add, or remapped to Resync), and the user will have to manually work with finalizers to better understand them (as well as use finalizers to make sure delete events aren't missed during downtime), so I think there's still some value in having the "Opinionated" reconciler that handles finalizer logic, just like with the watcher.

operator/reconciler.go
IfSentient: We can't wrap the interface here, as we're passing a`TypedReconcileRequest`, rather than a `ReconcileRequest` that uses `resource.Object`, which is what the `Reconciler` interface expects.

operator/reconciler.go
IfSentient: Yeah, this was mainly me rabbit-holing myself in edge cases where the reconciler successfully calls a third-party service, but then calls to update the object's status subresource fail for a transient reason (i.e. 5xx), so need to be retried in some exponential backoff way. It _does_ introduce a lot of complexity, so I think we can remove it if it doesn't get any usage.

ferruvich (approved):
LGTM - we already discussed about the `Status` field and it's ok to merge it 👍🏼

operator/reconciler.go
radiohead: Ah, good point – I've missed that. Let's keep it the way it is then 👍

operator/reconciler.go
radiohead: OK, since we have a plan to have this logic in the informer, let's keep this as is 👍

operator/reconciler.go
radiohead: I generally prefer to only err on the side of not adding things which might not end up being used, but I don't have a strong objection in this particular case.

radiohead (approved):

joeblubaugh:
@cesararmada - might be interesting as we think about multi-tenant backends. Would this simplify existing controller code?

Issue #49, opened 2023-07-04
Support upsert and force-delete logic in stores


radiohead:
### What

Add new methods (or options to existing methods) to support the following operations:

* Upsert – to support create-or-update logic. A client attempts to create a new resource and if it already exists it updates it with override to force a specific spec to exist in the API.
* Force-delete – to support use-cases where client wants to ensure that a certain resource is not present in the API, i.e. delete and ignore any possible 404 errors

### Why

Those operations are common when writing controller loops which create / modify downstream resources in response to upstream changes and right now we are forced to write this logic ourselves, which causes a lot of boilerplate code.

Issue #19, opened 2023-06-08
`resource` package should have sentinel errors


joeblubaugh:
`resource` has an interface `Client` that is used in several parts of the package. Errors from this client are returned directly to callers. In the case of a k8s client (the only implementation), these are k8s client or server errors and need to be processed as such.

In the case of another Client implementation, the type of errors returned by the Client will change. This means that app code calling a Client needs to process errors differently depending on the configured client, and the errors from these packages essentially become part of the SDKs interface contract.

I think some common sentinel errors might be useful, especially `Unauthorized`, `NotFound`, and `TooManyRequests`

radiohead:
Should be fixed by https://github.com/grafana/grafana-app-sdk/pull/730, we now properly parse & propagate k8s status errors, so one can use https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors for properly interpreting API server errors.

PR #57, opened 2023-07-17
Upsert and ForceDelete implementation for Store & TypedStore

mdefanti:
Implements #49

### DIFF
--- a/resource/store.go
+++ b/resource/store.go
@@ -217,7 +217,7 @@ func (s *Store) Delete(ctx context.Context, kind string, identifier Identifier) 
 	return client.Delete(ctx, identifier)
 }
 
-// Delete deletes a resource with the given Identifier and kind, ignores client 404 errors.
+// ForceDelete deletes a resource with the given Identifier and kind, ignores client 404 errors.
 func (s *Store) ForceDelete(ctx context.Context, kind string, identifier Identifier) error {
 	client, err := s.getClient(kind)
 	if err != nil {
--- a/resource/typedstore.go
+++ b/resource/typedstore.go
@@ -142,7 +142,7 @@ func (t *TypedStore[T]) Delete(ctx context.Context, identifier Identifier) error
 	return t.client.Delete(ctx, identifier)
 }
 
-// Delete deletes a resource with the provided identifier
+// ForceDelete deletes a resource with the provided identifier, ignores 404 errors
 func (t *TypedStore[T]) ForceDelete(ctx context.Context, identifier Identifier) error {
 	err := t.client.Delete(ctx, identifier)
 


resource/store.go
```diff
@@ -154,6 +157,53 @@ func (s *Store) UpdateSubresource(
...
+
+	client, err := s.getClient(obj.StaticMetadata().Kind)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := client.Get(ctx, obj.StaticMetadata().Identifier())
+
+	if err != nil {
+		cast, ok := err.(*k8sErrors.ServerResponseError)
```
IfSentient, line 185: This dovetails a bit into https://github.com/grafana/grafana-app-sdk/issues/19, but I think it may be a better approach to define an interface like
```go
type APIServerResponseError interface {
    error
    StatusCode() int
}
```
And try casting to that to check the error status code, which allows us to avoid the dependency cycle and not move around the error types in the `k8s` package. Thoughts?

IfSentient (commented):
Looks good! Just one comment about handling the kubernetes errors

resource/store.go
```diff
@@ -164,6 +217,21 @@ func (s *Store) Delete(ctx context.Context, kind string, identifier Identifier)
 	return client.Delete(ctx, identifier)
 }
 
+// Delete deletes a resource with the given Identifier and kind, ignores client 404 errors.
```
IfSentient, line 220: `Delete` -> `ForceDelete` (linter doesn't check these anymore)

resource/typedstore.go
```diff
@@ -105,6 +142,17 @@ func (t *TypedStore[T]) Delete(ctx context.Context, identifier Identifier) error
 	return t.client.Delete(ctx, identifier)
 }
 
+// Delete deletes a resource with the provided identifier
```
IfSentient, line 145: Same here, `Delete` -> `ForceDelete`

IfSentient (approved):
Looks good, just a couple minor godoc nitpicks.

resource/store.go
```diff
@@ -154,6 +160,53 @@ func (s *Store) UpdateSubresource(
...
+			return nil, err
+		}
+	}
+
+	if resp != nil {
+		return client.Update(ctx, Identifier{
+			Namespace: obj.StaticMetadata().Namespace,
+			Name:      obj.StaticMetadata().Name,
+		}, obj, UpdateOptions{
+			ResourceVersion: obj.CommonMetadata().ResourceVersion,
```
radiohead, line 201: @IfSentient @mdefanti this should actually be the version from the response, not the one from the object passed into the Upsert.

resource/store.go
```diff
@@ -154,6 +160,53 @@ func (s *Store) UpdateSubresource(
...
+		})
+	}
+	return client.Create(ctx, Identifier{
+		Namespace: obj.StaticMetadata().Namespace,
+		Name:      obj.StaticMetadata().Name,
+	}, obj, CreateOptions{})
```
radiohead, line 207: @IfSentient @mdefanti does this work if the passed object has a resourceVersion set? Regular k8s client doesn't allow this and will fail with an error.

PR #75, opened 2023-08-21
Tracing Instrumentation & Codegen Inclusion

IfSentient:
# Traces Instrumentation

## Library Changes

Added instrumentation for OpenTelemetry traces to the SDK, specifically the `operator` and `k8s` packages. Each package now has two extra exported functions (defined in `*/telemetry.go`):
* `SetTracer(t trace.Tracer)`, which sets the OTel Tracer used by the package when creating spans, and
* `GetTracer() trace.Tracer`, which returns the OTel Tracer set by GetTracer, _or_, if no Tracer has been set, uses the global `TracerProvider` from `otel.GetTracerProvider()` to create a Tracer with the package name (if no global TracerProvider has been set, OTel will give us a TracerProvider which returns dummy Tracers which do nothing, effectively disabling traces for the package until and unless `SetTracer` is called).

Contexts which are propagated to user code (such as in the `operator` package) contain span information so that tracing can be continued into the app code.

## Local Environment Changes

* `tempo` has been added as registered data source in the local `config.yaml`
* The ini configmap for grafana has been updated to turn on tracing and point at the tempo service. This is required for the plugin tracing to work.

## Codegen Changes

Changes have been made in the generated project code created by `grafana-app-sdk project component add operator`, which now generates instrumentation set-up for OpenTelemetry tracing, and adds span generation to the dummy watcher functions.

Generated plugin code for `grafana-app-sdk prokect component add backend` has been updated to use `app.Manage` for `plugin/pkg/main.go`, [which auto-orchestrates traces](https://grafana.com/docs/grafana/latest/developers/plugins/create-a-grafana-plugin/extend-a-plugin/add-distributed-tracing-for-backend-plugins/), and spans have also been added to the generated handlers.


marefr:
If you want to spice up things even further you might be able to use the httptrace/otelhttptrace package to get low level transport/connection details, example https://github.com/grafana/grafana/blob/d1218db0230888f98769bdd2aca2797af0646478/pkg/infra/httpclient/httpclientprovider/tracing_middleware.go#L30

k8s/gvclient.go
```diff
@@ -32,26 +34,50 @@ type groupVersionClient struct {
...
 	bytes, err := request.Do(ctx).StatusCode(&sc).Raw()
+	span.SetAttributes(
+		attribute.Int("http.response.status_code", sc),
+		attribute.String("http.request.method", http.MethodGet),
+		attribute.String("server.address", request.URL().Hostname()),
+		attribute.String("server.port", request.URL().Port()),
+		attribute.String("url.full", request.URL().String()),
+	)
 	if err != nil {
 		return nil, parseKubernetesError(bytes, sc, err)
```
onprem, line 82: Do we want to set the span status to `error` here? Similar to how it's done for `*groupVersionClient.get()` here https://github.com/grafana/grafana-app-sdk/blob/7caf8e0298fcb2b760914c3b5b9fe76304b479f9/k8s/gvclient.go#L54

operator/informer_controller.go
```diff
@@ -277,6 +283,8 @@ func (c *InformerController) informerUpdateFunc(resourceKind string) func(contex
 			}
 			if err != nil && c.RetryPolicy != nil {
 				c.queueRetry(retryKey, err, func() (*time.Duration, error) {
+					ctx, span := GetTracer().Start(ctx, "controller-retry")
```
onprem, line 286: Just some thoughts on this. Between having a separate span for retry, under the same trace like this vs aggregating all retries under one big span, I think I'd prefer having separate spans, as it feels more correct. As I understand, the retry span starts when the retry actually gets triggered from queue, which seems the right way of going about it.

onprem (approved):
Great work! This looks good to me. Just a small nit in the comments.

k8s/gvclient.go
```diff
@@ -32,26 +34,50 @@ type groupVersionClient struct {
...
 	if strings.TrimSpace(identifier.Namespace) != "" {
 		request = request.Namespace(identifier.Namespace)
 	}
 	bytes, err := request.Do(ctx).StatusCode(&sc).Raw()
+	span.SetAttributes(
+		attribute.Int("http.response.status_code", sc),
+		attribute.String("http.request.method", http.MethodGet),
+		attribute.String("server.address", request.URL().Hostname()),
+		attribute.String("server.port", request.URL().Port()),
+		attribute.String("url.full", request.URL().String()),
```
radiohead, line 50: Shouldn't this be part of the http.request attributes?
```suggestion
		attribute.String("http.request.url", request.URL().String()),
```

k8s/gvclient.go
```diff
@@ -32,26 +34,50 @@ type groupVersionClient struct {
...
 	request := g.client.Get().Resource(plural).Name(identifier.Name)
 	if strings.TrimSpace(identifier.Namespace) != "" {
 		request = request.Namespace(identifier.Namespace)
 	}
 	bytes, err := request.Do(ctx).StatusCode(&sc).Raw()
+	span.SetAttributes(
+		attribute.Int("http.response.status_code", sc),
+		attribute.String("http.request.method", http.MethodGet),
+		attribute.String("server.address", request.URL().Hostname()),
+		attribute.String("server.port", request.URL().Port()),
```
radiohead, line 49: Do we need these? They are already in the URL and I don't think I personally would be searching / grouping by these.

k8s/telemetry.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
+)
+
+var tracer trace.Tracer
+
+// SetTracer sets the tracer used for generating spans for this package
+func SetTracer(t trace.Tracer) {
+	tracer = t
```
radiohead, line 12: NIT – these can cause data races. I don't think the probability is high, but the impact is a panic, so maybe worth adding an rwmutex here.

operator/telemetry.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
+)
+
+var tracer trace.Tracer
+
+// SetTracer sets the tracer used for generating spans for this package
+func SetTracer(t trace.Tracer) {
+	tracer = t
```
radiohead, line 12: Same as in `k8s` package, might be worth adding a mutex here.

codegen/templates/plugin/handler_models.tmpl
```diff
@@ -22,23 +24,27 @@ type {{.Name}}Service interface {
 }
 
 func (p *Plugin) handle{{.Name}}List(ctx context.Context, req router.JSONRequest) (router.JSONResponse, error) {
+	ctx, span := tracing.DefaultTracer().Start(ctx, "{{.MachineName}}-list")
+	defer span.End()
```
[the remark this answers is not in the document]
IfSentient: Yeah, I started writing a tracing middleware to also include more info in span attributes, then basically realized I wasn't sure what the span should be, and that _also_ grafana includes a span for the actual CallResource call that gets sent to the plugin, so it seemed like redundant info. I think it may still be useful to have an optional additional tracing middleware, perhaps in a subsequent PR.

k8s/gvclient.go
IfSentient: Per the [Semantic Conventions for OpenTelemetry HTTP Spans](https://github.com/open-telemetry/semantic-conventions/blob/098db1ca510da01fe941f6c6308aad5000def910/docs/http/http-spans.md#http-client), they just list `url.full` (not 100% sure why a lot of them don't have the `http` prefix, but was trying to abide by them).

k8s/gvclient.go
IfSentient: `server.address` is listed as required in the [Semantic Conventions for HTTP Spans](https://github.com/open-telemetry/semantic-conventions/blob/098db1ca510da01fe941f6c6308aad5000def910/docs/http/http-spans.md#http-client), and `server.port` is required if non-default for the protocol, which I didn't want to have a check each time for.

k8s/gvclient.go
```diff
@@ -66,12 +92,21 @@ func (g *groupVersionClient) getMetadata(ctx context.Context, identifier resourc
...
 		request = request.Namespace(identifier.Namespace)
 	}
 	err := request.Do(ctx).StatusCode(&sc).Error()
+	span.SetAttributes(
+		attribute.Int("http.response.status_code", sc),
+		attribute.String("http.request.method", http.MethodGet),
+		attribute.String("server.address", request.URL().Hostname()),
+		attribute.String("server.port", request.URL().Port()),
+		attribute.String("url.full", request.URL().String()),
+	)
```
ferruvich, line 112: Same as mentioned by Prem above, do we want to set span status to `error`?

ferruvich (approved):
Nice work mate! 
Added some comments on top of the things Prem and Igor mentioned

codegen/templates/plugin/handler_models.tmpl
```diff
@@ -22,23 +24,27 @@ type {{.Name}}Service interface {
...
 	filtersRaw := req.URL.Query().Get("filters")
 	filters := make([]string, 0)
 	if len(filtersRaw) > 0 {
 		filters = strings.Split(filtersRaw, ",")
 	}
 	svc, err := p.service.Get{{.Name}}Service(ctx)
 	if err != nil {
-	    log.DefaultLogger.Error("Error getting {{.Name}}Service: " + err.Error())
+	    log.DefaultLogger.Error("Error getting {{.Name}}Service: " + err.Error(), "error", err, "traceID", span.SpanContext().TraceID())
```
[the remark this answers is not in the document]
IfSentient: Yeah, I was holding off until the next PR with logging, as we'll no longer be using the `log.DefaultLogger` for things in the codegen, but 'll make this change as it's mainly a semantic one 👍

joeblubaugh (commented):
Any reason not to add this to the plugin/router types as well?

k8s/gvclient.go
radiohead: Aha, good to know. Thanks!

k8s/gvclient.go
radiohead: Good to know. Thanks!

radiohead (approved):

IfSentient:
> Any reason not to add this to the plugin/router types as well?

My main reasoning was just that it seemed like grafana's spans already contained all the information we could include in the plugin/router, but if you think that would be valuable, I can go ahead and add tracing there, too. I'll merge this PR and create a separate one for adding tracing to the router(s) so we can work out exactly what would be useful to have there in that one.

PR #84, opened 2023-08-29
Chore: move kindsys into app-sdk and simplify

ryantxu:
This moves a copy of kindsys from the currently used branch directly into this repository, and removes as much unused stuff as possible.

This will break the coupling to kindsys -- and focus on more generic k8s resource shapes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,12 +8,12 @@ require (
 	github.com/grafana/codejen v0.0.3
 	github.com/grafana/cuetsy v0.1.11
 	github.com/grafana/grafana-plugin-sdk-go v0.173.0
-	github.com/grafana/kindsys v0.0.0-20230508162304-452481b63482
 	github.com/grafana/thema v0.0.0-20230511182720-3146087fcc26
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/puzpuzpuz/xsync/v2 v2.5.0
 	github.com/spf13/cobra v1.7.0
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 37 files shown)


go.mod
```diff
@@ -25,6 +25,7 @@ require (
 )
 
 require (
+	cloud.google.com/go/compute/metadata v0.2.3 // indirect
```
ryantxu, line 28: I added this with:
```
go get cloud.google.com/go/compute/metadata
```
because `go mod tidy` would not resolve without it

go.mod
```diff
@@ -8,12 +8,12 @@ require (
 	github.com/grafana/codejen v0.0.3
 	github.com/grafana/cuetsy v0.1.11
 	github.com/grafana/grafana-plugin-sdk-go v0.173.0
-	github.com/grafana/kindsys v0.0.0-20230508162304-452481b63482
```
ryantxu, line 11: ^^^ the real change!

kindsys/cue.mod/module.cue
```diff
@@ -0,0 +1 @@
+module: "github.com/grafana/grafana-app-sdk/kindsys"
```
ryantxu, line 1: if anything is broken that is not exercised by these tests, i suspect the package name above is the cause

go.mod
IfSentient: 👍 I had noticed the issue, but doing `go get` on my end didn't seem to actually modify the `go.mod`, just made `go mod tidy` start working again.

PR #86, opened 2023-09-01
[router] Tracing Middleware

IfSentient:
Added tracing middleware to the `router` package, which wraps every request in a new span and adds all HTTP attributes to the span.

This is a follow-up to the PR to add tracing, CC @joeblubaugh


plugin/router/middleware.go
```diff
@@ -54,3 +57,30 @@ func NewCapturingMiddleware(f func(ctx context.Context, r *backend.CallResourceR
...
+		ctx, span := tracer.Start(ctx, "middleware")
+		defer span.End()
+		routeInfo := MatchedRouteFromContext(ctx)
+
+		resp := next(ctx)
+
+		query := ""
+		if s := strings.SplitN(req.URL, "?", 1); len(s) > 1 {
+			query = s[1]
+		}
```
radiohead, line 76: Why not use `req.URL.RawQuery`?

radiohead (approved):

ferruvich (approved):
LGTM 👍🏼 agreeing with Igor's comment about the usage of `RawQuery`

radiohead (approved):

plugin/router/middleware.go
radiohead: Ah, of course! I completely forgot that we're not dealing with an `http.Request` here but rather a custom one.

joeblubaugh: was there any consideration of redacting information that's present in the query string?

Issue #24, opened 2023-06-09
Logs, Metrics, Tracing


IfSentient:
An issue of two parts:
1. The library code (all packages that aren't `cmd/*` and `codegen`) should have hooks for accepting interfaces for loggers, metrics, and tracing. It should be simple for a user of the SDK to get logs or metrics from underlying systems, such as the kubernetes client, informers, or OpinionatedWatchers.
2. The boilerplate code generated with `grafana-app-sdk project component add` should include pre-built solutions for logging, metrics, and traces.
```[tasklist]
### Tasks
- [x] Add Tracing Instrumentation to the SDK (https://github.com/grafana/grafana-app-sdk/pull/75)
- [x] Add Logging Instrumentation to the SDK (https://github.com/grafana/grafana-app-sdk/pull/89)
- [x] Add Metric instrumentation to the SDK (https://github.com/grafana/grafana-app-sdk/pull/103)
- [ ] Update Docs
```

ellisda:
FWIW, the SLO app was interested in metrics for our telemetry (i.e. counter metrics of how many times our users CRUD their SLOs). see https://github.com/grafana/product_analytics/issues/54. If our api/controller weren't running inside the hosted-grafana pod, we could emit our own metrics but we don't have away to get our metrics scraped at present.

When we start running a multitenant controller, outside of hosted-grafana pod, we'll be able to emit the metrics ourselves, but we had wondered if there was some piece of infrastructure in the CRD api server that might be able to expose these kinds of counters for us (and for other kinds, rather than have each kind's controller do it from scratch).

I'm not sure this request is in-scope for your issue #24 here, but thought I'd elaborate.

radiohead:
Marking this as closed since the SDK has been instrumented.

PR #89, opened 2023-09-06
Logging

IfSentient:
# What this PR does
This PR adds the `logging` package to the repo, which contains a `Logger` interface definition, an `slog` implementation of that interface, methods to put a Logger into the context/retrieve it from the context, and a DefaultLogger which is used when either references directly, or no Logger is in the interface (defaults to a No-op logger).

It also introduces using logging in a few places, and a `router.NewLoggingMiddleware` which logs each request, and `plugin.PluginLogger`, which wraps a plugin-sdk-go `log.Logger` to be compliant with `logging.Logger`, and add `traceID` to each message logged with a context.

Generated project code now initializes the DefaultLogger in the plugin to a `plugin.PluginLogger` wrapping plugin-sdk-go's `log.DefaultLogger` (with a `pluginID` field attached), and in the operator to the slog implementation.

Local env setup now allows for specifying a `loki` data source, which gets set-up to monitor pod logs in the local kubernetes cluster.

## Breaking Changes
* `operator.InformerController.ErrorHandler` function now takes both a `context.Context` and an `error`, as opposed to previously just taking an `error`. The default handler will now log the error.
* `router.NewJSONRouter` and `router.NewJSONRouterWithErrorHandler` no longer take a `log.Logger` parameter, and will use the `logging.Logger` from the context (or `logging.DefaultLogger`) to log errors.

## Other Updates
The golang version in `go.mod` has been updated to 1.21, as that introduced `log/slog` to the standard library.

The golangci linter needed to be updated to handle go 1.21. Once updated, a few minor updates had to be made to unrelated files based on new enforcement (namely `interface{}` -> `any`). The GH action linter has also been updated, but I'm not sure if it will run the new one on the PR branch.

Part of https://github.com/grafana/grafana-app-sdk/issues/24


logging/slog.go
```diff
@@ -0,0 +1,64 @@
...
+func NewSLogLogger(handler slog.Handler) *SLogLogger {
+	return &SLogLogger{
+		Logger: *slog.New(&traceIDHandler{next: handler}),
+	}
+}
+
+// SLogLogger wraps slog.Logger to override the With() method to return a Logger (*SLogLogger), rather than *slog.Logger,
+// implementing the Logger interface.
+type SLogLogger struct {
+	slog.Logger
```
radiohead, line 26: Not sure if it's a good idea to wrap a value here, how about we wrap a pointer – this will also make New and With less awkward.
```suggestion
	*slog.Logger
```

logging/logger.go
```diff
@@ -0,0 +1,134 @@
...
+	DefaultLogger Logger = &NoOpLogger{}
+
+	contextKey = loggerContextKey{}
+)
+
+type loggerContextKey struct{}
+
+// FromContext returns the Logger set in the context with Context(), or the DefaultLogger if no Logger is set in the context.
+// If DefaultLogger is nil, it returns a *NoOpLogger so that the return is always valid to call methods on without nil-checking.
+func FromContext(ctx context.Context) Logger {
```
radiohead, line 18: Seeing as this is going to be the usual pattern of using the logger I suggest adding the context to the logger here – this will make it so that calls like `logging.FromContext(ctx).Info(...)` will always carry the context down to e.g. `traceIDHandler` as opposed to requiring the user to call `logging.FromContext(ctx).InfoContext(...)`, especially since a user might not necessarily be aware or care about the latter invocation.

This means in practice that we'd need to add a new method to the logger interface, e.g. `WithContext` or similar. Interestingly enough, `slog` used to have such method at some point but it seems like they got rid of it upon release.

plugin/router/json_router.go
```diff
@@ -78,20 +78,18 @@ type JSONErrorHandler func(err plugin.Error) (int, JSONResponse)
 type JSONRouter struct {
 	*Subrouter
 	errHandler JSONErrorHandler
-	logger     log.Logger
 }
 
 // NewJSONRouter returns a new JSONRouter with a logger.
-func NewJSONRouter(logger log.Logger) *JSONRouter {
-	return NewJSONRouterWithErrorHandler(logger, nil)
+func NewJSONRouter() *JSONRouter {
```
radiohead, line 84: I don't think we have to solve this problem immediately in this PR but I'd like to point out a problem of knowing where in code a log line has come from. Previously I've been hand-rolling a combination of a required "component" field and merging logging attributes between a default component logger and a logger from context – I wonder if we could automate this somehow. One option would be to use something like `runtime.Caller`, another to keep passing loggers to initialisers and attaching a `component` field to them.

plugin/router/middleware.go
```diff
@@ -59,6 +62,21 @@ func NewCapturingMiddleware(f func(ctx context.Context, r *backend.CallResourceR
...
+// and injects the provided Logger into the context used downstream.
+func NewLoggingMiddleware(logger logging.Logger) MiddlewareFunc {
+	return func(handler HandlerFunc) HandlerFunc {
+		return func(ctx context.Context, req *backend.CallResourceRequest, sender backend.CallResourceResponseSender) {
+			logger.InfoContext(ctx, fmt.Sprintf("Handling %s request to path %s", req.Method, req.Path),
+				"request.http.method", req.Method,
+				"request.http.path", req.Path,
+				"request.user", req.PluginContext.User.Name)
+
+			handler(logging.Context(ctx, logger), req, sender)
```
radiohead, line 75: NIT
```suggestion
			start := time.Now()
			handler(logging.Context(ctx, logger), req, sender)
			lat := time.Since(start)
			// Logging latency in ms because it's easier to perceive as a human.
			// But we also attach a separate field where latency is in seconds, for e.g. Loki queries.
			logger.InfoContext(ctx, fmt.Sprintf("%s %s %dms", req.Method, req.Path, lat.Milliseconds()),
				"request.http.method", req.Method,
				"request.http.path", req.Path,
				"request.user", req.PluginContext.User.Name,
				"request.latency", lat.Seconds(),
			)
```

logging/logger.go
IfSentient: Do we think it may be useful to drop the `<X>Context` methods from `Logger` and instead just have `WithContext(context.Context) Logger` which would ultimately achieve the same effect? Or is there value in having both?

plugin/router/json_router.go
IfSentient: True, I also wanted to make sure that if a Logger is set upstream in the request context, the one used for error logging by the router is the one set in the context, rather than a different one. Probably good to add some additional standard fields that can be attached with `With` in both the router calls and in upstream logging.

logging/logger.go
radiohead: > Do we think it may be useful to drop the `<X>Context` methods from `Logger` and instead just have `WithContext(context.Context) Logger` which would ultimately achieve the same effect? Or is there value in having both?

I am in favour of dropping the extra methods and enforcing the `WithContext(...).Info(...)` approach.

codegen/templates/operator/watcher.tmpl
radiohead, line 1: @IfSentient I think we should not be passing context to `Debug(...)` calls in this file, possibly a leftover from the old version of the API?

codegen/templates/operator/watcher.tmpl
```diff
@@ -30,7 +31,7 @@ func (s *{{.Name}}Watcher) Add(ctx context.Context, rObj resource.Object) error
     }
 
     // TODO
-    fmt.Println("Added ", object.StaticMetadata().Identifier().Name, span.SpanContext().TraceID())
+    logging.FromContext(ctx).Debug(ctx, "Added resource", "name", object.StaticMetadata().Identifier().Name)
```
radiohead, line 34: e.g. this should be
```suggestion
    logging.FromContext(ctx).Debug("Added resource", "name", object.StaticMetadata().Identifier().Name)
```

plugin/router/json_router.go
radiohead: Makes sense to me – I think we can address that in a follow-up issue / PR 👍

plugin/logging.go
```diff
@@ -0,0 +1,64 @@
...
+
+// WarnContext adds the traceID field to the underlying log.Logger, then calls Warn with the provided msg and args
+func (p *PluginLogger) WarnContext(ctx context.Context, msg string, args ...any) {
+	p.Logger.With(logging.TraceIDKey, trace.SpanContextFromContext(ctx).TraceID()).Warn(msg, args...)
+}
+
+// ErrorContext adds the traceID field to the underlying log.Logger, then calls Error with the provided msg and args
+func (p *PluginLogger) ErrorContext(ctx context.Context, msg string, args ...any) {
+	p.Logger.With(logging.TraceIDKey, trace.SpanContextFromContext(ctx).TraceID()).Error(msg, args...)
+}
```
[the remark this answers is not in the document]
IfSentient: Whoops, I thought I fixed this when I changed the Logger interface, but it looks like that file change didn't get commited.

codegen/templates/operator/watcher.tmpl
radiohead: Looks like there are still a few more instances of those problematic calls – I left a few suggestions to fix those 🙂

codegen/templates/operator/watcher.tmpl
```diff
@@ -66,7 +67,7 @@ func (s *{{.Name}}Watcher) Delete(ctx context.Context, rObj resource.Object) err
     }
 
     // TODO
-    fmt.Println("Deleted ", object.StaticMetadata().Identifier(), span.SpanContext().TraceID())
+    logging.FromContext(ctx).Debug(ctx, "Deleted resource", "name", object.StaticMetadata().Identifier().Name)
```
radiohead, line 70:
```suggestion
    logging.FromContext(ctx).Debug("Deleted resource", "name", object.StaticMetadata().Identifier().Name)
```

plugin/logging.go
```diff
@@ -0,0 +1,64 @@
...
+// InfoContext, WarnContext, or ErrorContext are called.
+// nolint:revive
+type PluginLogger struct {
+	log.Logger
+	ctx context.Context
+}
+
+// Debug adds the traceID field to the underlying log.Logger from the context embedded with WithContect, then calls Debug with the provided msg and args
+func (p *PluginLogger) Debug(msg string, args ...any) {
+	p.Logger.With(logging.TraceIDKey, trace.SpanContextFromContext(p.ctx).TraceID()).Debug(msg, args...)
```
radiohead, line 30: Should we check the presence of span in the context here? Otherwise I think we might get some logs with empty traceID fields.

IfSentient: Yup, makes sense, I was trying to maximize for efficiency, but probably not the best place to do so.

radiohead (approved):

codegen/templates/plugin/plugin.tmpl
```diff
@@ -38,13 +38,14 @@ func (p *Plugin) CallResource(ctx context.Context, req *backend.CallResourceRequ
...
 	p := &Plugin{
-	    router: router.NewJSONRouter(log.DefaultLogger),
+	    router: router.NewJSONRouter(),
 			namespace: namespace,
 	    service: service,
 	}
 
 	p.router.Use(
 		router.NewTracingMiddleware(otel.GetTracerProvider().Tracer("tracing-middleware")),
+		router.NewLoggingMiddleware(logging.DefaultLogger),
```
ferruvich, line 48: Do you think it's better this logging middleware consumes the DefaultLogger or should we let `New` consume the logger used in `newInstanceFactory` and pass it here too?
From what I can see the `newInstanceFactory` might be a good place to set some additional fields from instance settings. Or do you think that `DefaultLogger` would be the one having those fields?

logging/logger.go
```diff
@@ -0,0 +1,77 @@
...
+	DefaultLogger Logger = &NoOpLogger{}
+
+	contextKey = loggerContextKey{}
+)
+
+type loggerContextKey struct{}
+
+// FromContext returns the Logger set in the context with Context(), or the DefaultLogger if no Logger is set in the context.
+// If DefaultLogger is nil, it returns a *NoOpLogger so that the return is always valid to call methods on without nil-checking.
+// So long as the Logger to return is not the NoOpLogger, it will call WithContext(ctx) on the logger in the context and return
```
ferruvich, line 18: Do you think this could be rephrased to something like "So long as the Logger to return is not the NoOpLogger, it will always carry down the input context" (probably there's a much better way to phrase it)?
I believe this is more direct and doesn't require people to scroll down to `Logger` interface to check what `WithContext` function does.

logging/logger.go
```diff
@@ -0,0 +1,77 @@
...
+	return &NoOpLogger{}
+}
+
+// Context returns a new context built from the provided context with the provided logger in it.
+// The Logger added with Context() can be retrieved with FromContext()
+func Context(ctx context.Context, logger Logger) context.Context {
+	return context.WithValue(ctx, contextKey, logger)
+}
+
+// Logger extends the BasicLogger interface with methods that accept a context as well as the message and key/val args.
```
ferruvich, line 39: Should this be rephrased as:
- I can't see `BasicLogger`
- Methods carrying context are not there anymore

ferruvich (approved):
Sorry for being late on this!
LGTM 👍🏼 left some comments

logging/logger.go
IfSentient: Good call, I'll update the godocs!

codegen/templates/plugin/plugin.tmpl
IfSentient: In this case, I guess I mostly assumed that the default logger for the plugin would have all the expected information from the plugin init process.

ferruvich: Is fine for me to assume that as the default logger can be updated in the `app.Manage`  instance Factory. 
The important thing here is that is clear to developers that they can do it 👍🏼

ferruvich (approved):
LGTM 👍🏼 nice work!

PR #90, opened 2023-09-06
[WIP/Proposal] `simple` Package

IfSentient:
This is some very WIP code and ideas for a `simple` package, which allows a user to build apps without needing much from the `grafana-app-sdk project component add` command, instead being able to do something like:

```go
func main() {
    // ...load kube config, webhook config (TLS)...
    operator, err := simple.NewOperator(kubeConfig, simple.WebhookConfig{
        Enabled: true,
        Port: 12345,
        TLSConfig: loadedTLSConfig,
    })
    // Add a reconciler (wrapped in OpinionatedReconciler) for a kind, along with an informer
    operator.ReconcileKind(mykind.Schema(), &simple.Reconciler{
        ReconcileFunc: func(ctx context.Context, req operator.ReconcileRequest) (operator.ReconcileResult, error) {
            // TODO, example-only, reconciler would likely be defined elsewhere
        }}, simple.ListWatchOptions{})
    // Add a watcher (wrapped in OpinionatedWatcher) for a kind, along with an informer
    operator.WatchKind(myotherkind.Schema(), &simple.Watcher{}, simple.ListWatchOptions{})
    // Validation for a kind
    operator.ValidateKind(mykind.Schema(), someValidator)
    // Mutate a kind on ingress
    operator.MutateKind(myotherkind.Schema(), someMutator)

    // Run the whole thing
    operator.Run(closeChan)
}
```
and similar for a plugin (looking to eventually incorporate the plugin-sdk-go's `app.Manage` so that written code becomes as close to "just business logic" as possible).

This PR is mainly looking to gather thoughts and opinions on this kind of package and whether it's worth going forward past this rough outline/POC


simple/watcher.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+
+	"github.com/grafana/grafana-app-sdk/resource"
+)
+
+// Watcher is a struct that implements operator.ResourceWatcher and SyncWatcher, but takes no action on its own.
+// For each method in (Add, Update, Delete) the corresponding exported function field is called, if non-nil.
+type Watcher struct {
```
radiohead, line 11: What would be the use-case for this?

simple/operator.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/grafana/grafana-app-sdk/operator"
+	"github.com/grafana/grafana-app-sdk/resource"
+	"k8s.io/client-go/rest"
+)
+
+type WebhookConfig struct {
+	Enabled   bool
+	Port      int
+	TLSConfig k8s.TLSConfig
+}
```
radiohead, line 18: I would add validating and mutating controllers as part of this configuration. I don't see a reason why someone would enable webhooks but leave out controllers and vice versa. This would also simplify the code.

radiohead (commented):
I like the idea but I wonder if we could start with something even more limited, e.g. use the existing `ResourceGroupRouter` and make sure we use typed primitives as much as possible.

simple/watcher.go
IfSentient: Namely to be easily wrapped by `operator.OpinionatedWatcher` with a dedicated `Sync` method to map to the `OpinionatedWatcher` sync events.

radiohead: Sorry, my bad – I wasn't clear enough with my question. IIUC this is used to provide the `WatchKind` functionality, but I was wondering whether we even should provide it.

I suggest that we limit the simple use-case to reconciling operators only, WDYT?

PR #92, opened 2023-09-11
Add Generation to CommonMetadata

IfSentient:
Move the `generation` field from `resource.CommonMetadata.ExtraFields["generation"]` to `resource.CommonMetadata.Generation`. Ensure that generated code from the SDK handles this move properly, and that the translation code in the `k8s` package does as well.


radiohead (approved):
:shipit:

PR #105, opened 2023-09-20
[resource] Stores Update() methods update CommonMetadata.UpdateTimestamp

IfSentient:
Have resource-package Stores Update() methods update CommonMetadata.UpdateTimestamp in the object they're updating. This is a semi-temporary measure to prevent needing admission control to update the UpdateTimestamp, but also keep the kubernetes client from taking on what should be business logic.

### DIFF
--- a/resource/store.go
+++ b/resource/store.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 )
 
 // TODO: rewrite the godocs, this is all copied from crd/store.go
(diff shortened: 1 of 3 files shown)


resource/store.go
```diff
@@ -122,6 +123,10 @@ func (s *Store) Update(ctx context.Context, obj Object) (Object, error) {
 		return nil, fmt.Errorf("obj.StaticMetadata().Name must not be empty")
 	}
 
+	md := obj.CommonMetadata()
+	md.UpdateTimestamp = time.Now().UTC()
```
ellisda, line 127: While I can think of uses in our SLO app, where we might have (mis)used the capability to set this timestamp, I like this approach better where the SDK or the infra behind it takes ownership of these timestamps.

FWIW, we're interested in measuring end to end latency, and so we might have (mis)used the ability to set the "update" timestamp to when our user's initiated the update (i.e. client-side timestamp, maybe from the javascript browser code). This would feel dangerous, though, clock skew from different machines, user able to put in obviously invalid time, etc.

radiohead (approved):

PR #116, opened 2023-10-06
CRD Conversion Webhooks

IfSentient:
This PR introduces the ability to add a CRD conversion webhook to the `k8s.WebhookServer`. A user can supply any number of Group/Kind combinations with associated converters, described by an interface. Local dev environment config and manifest generation has also been updated to handle CRD conversion, now allowing for an additional bit of config in the webhook options.


k8s/webhooks.go
```diff
@@ -299,6 +325,82 @@ func (w *WebhookServer) HandleMutateHTTP(writer http.ResponseWriter, req *http.R
...
+		}, rev.Request.DesiredAPIVersion)
+		if err != nil {
+			// Conversion error
+			writer.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		rev.Response.ConvertedObjects = append(rev.Response.ConvertedObjects, runtime.RawExtension{
+			Raw: res,
+		})
+	}
```
radiohead, line 393: I don't think we should be returning non-200 responses here, instead we should use the `Response.Result` accordingly, since at this point all errors are related to the "business logic", not the transport protocol.

I suggest copying the logic from [the Kubernetes conversion test](https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/crd-conversion-webhook/converter/framework.go#L88-L115) into here.

k8s/conversion.go
```diff
@@ -0,0 +1,25 @@
+package k8s
+
+// Converter describes a type which can convert a kubernetes kind from one API version to another.
+// Typically there is one converter per-kind, but a single converter can also handle multiple kinds.
+type Converter interface {
+	// Convert converts a raw kubernetes kind into the target APIVersion.
+	// The RawKind argument will contain kind information and the raw kubernetes object,
+	// and the returned bytes are expected to be a raw kubernetes object of the same kind and targetAPIVersion
+	// APIVersion.
```
radiohead, line 9: According to the webhook documentation there is another requirement which we should probably be passing on to the `Converter` implementations (since we don't handle it ourselves):
```
The webhook is expected to set `apiVersion` of these objects to the `request.desiredAPIVersion`.
```

We could also look into simplifying the interface and making sure that we set the version ourselves, in the webhook handler code.

IfSentient: Ah, I guess I wasn't explicit in my assumption that the returned object should have the correct `kind` and `apiVersion`. I'll make sure the godocs are explicit about this.

k8s/webhooks.go
```diff
@@ -355,23 +355,33 @@ func (w *WebhookServer) HandleConvertHTTP(writer http.ResponseWriter, req *http.
...
 	// Go through each object in the request
 	for _, obj := range rev.Request.Objects {
 		// Partly unmarshal to find the kind and APIVersion
 		tm := metav1.TypeMeta{}
 		err = json.Unmarshal(obj.Raw, &tm)
 		if err != nil {
-			writer.WriteHeader(http.StatusBadRequest)
-			return
+			rev.Response.Result.Status = metav1.StatusFailure
+			rev.Response.Result.Code = http.StatusInternalServerError
```
radiohead, line 371:
```suggestion
			rev.Response.Result.Code = http.StatusBadRequest
```

k8s/webhooks.go
```diff
@@ -355,23 +355,33 @@ func (w *WebhookServer) HandleConvertHTTP(writer http.ResponseWriter, req *http.
...
+			break
 		}
 		// Get the associated converter for this kind
 		conv, ok := w.converters[gk(tm.GroupVersionKind().Group, tm.Kind)]
 		if !ok {
 			// No converter for this kind
-			writer.WriteHeader(http.StatusInternalServerError)
-			return
+			rev.Response.Result.Status = metav1.StatusFailure
+			rev.Response.Result.Code = http.StatusInternalServerError
```
radiohead, line 381:
```suggestion
			rev.Response.Result.Code = http.StatusUnprocessableEntity
```

radiohead (approved):
A few nits about response codes but overall looks good! :shipit:

radiohead (approved):

PR #144, opened 2023-11-07
[codegen] Introduce new base interface for codegen, refactor jennies to remove thema dependency

IfSentient:
## What this PR does

This PR introduces the `codegen.Kind` interface as the accepted type for all codegen jennies, rather than the thema-centric `kindsys.Custom`. Thema-related code has been moved from `codegen` to `codegen/thema`, and generators have all been moved into `codegen/jennies` (excepting some thema-specific jennies which have to consume the lineage). Generators have been refactored to use `codegen.Kind`, and some pieces of `thema/internal` have been extracted into `codegen/jennies/dstutil.go` and `codegen/jennies/cuetil.go` to allow for go types codegen which produces the same output. Imports have been updated in `cmd` to account for these changes.

This PR looks a bit larger than it really is because git labels the moving and renaming (and then minor changes from `kindsys.Custom` to `codegen.Kind`) of all the non-CRD/non-GoTypes/non-TS jennies as new files, when in reality only a few lines in them have changed.

## Codegen Output Changes

Only two files in the generated output have any changes as a result of this PR, but their underlying behavior remains the same. The first is [the resource.Object implementation file](https://github.com/grafana/grafana-app-sdk/pull/144/files#diff-60a128dfee744046af33cb4f97cf4679f6f9aec231ba3057b1198074961734f5), which has a change to its `Unmarshal` method in preparation for generic codegen. It now defaults to using `json.Unmarshal` unless a new function is specified. To this end, [<kind>_lineage_gen.go](https://github.com/grafana/grafana-app-sdk/pull/144/files#diff-188b41dea51a3ed5783c8497efdc8ecaa87666eaed25113ea1e63e998e8de959) has been updated to set that function to the contents of `Unmarshal` previously, using the thema-specific code.

## Next steps

The next PR will introduce the vanilla CUE kind, and generic code generation with parsers (with updates to the CLI to allow for either thema or plain CUE input based on a flag).


radiohead (commented):
Sorry for taking so long to review this – it's rather big! I didn't spot anything obviously wrong but I might've missed something. Still, I think it should be OK to merge.

radiohead (approved):
Whoops, meant to approve instead of commenting. Here it goes!

PR #155, opened 2023-11-14
[codegen/CLI] Final Part of non-Thema Codegen

IfSentient:
This PR merges the final part of the non-thema-based codegen branch, adding thew new kind format definition and parser in `codegen/cuekind`, and updating the CLI to default to non-thema-based codegen, and using `-f` to allow the user to explicitly specify the format of their kinds (`-f` is allowed to be `cue` or `thema`, defaulting to `cue`). 

This is a breaking change, as the default behavior of the CLI is changing, but users upgrading should be able to retain full compatibility with their previous codegen by simply adding `-f thema` to their CLI commands.

The new format is defined in `codegen/cuekind/def.cue`. A subsequent PR will update the tutorial and other documentation which talks about thema to use the new CUE definition instead.


cmd/grafana-app-sdk/generate.go
```diff
@@ -194,3 +243,62 @@ func generateCRDs(parser *themagen.CustomKindParser, genPath string, encoding st
...
+		return nil, err
+	}
+	for i, f := range crdFiles {
+		crdFiles[i].RelativePath = filepath.Join(cfg.CRDPath, f.RelativePath)
+	}
+
+	allFiles := append(make(codejen.Files, 0), resourceFiles...)
+	allFiles = append(allFiles, modelFiles...)
+	allFiles = append(allFiles, tsFiles...)
+	allFiles = append(allFiles, tsFiles...)
```
radiohead, line 302: Looks like instead of appending CRD files this is appending TS files two times.
```suggestion
	allFiles = append(allFiles, crdFiles...)
```

radiohead (commented):
Looks good! Just a couple of small issues 🙂

ferruvich (commented):
LGTM! Agreeing on Igor's comments 😄

radiohead (approved):

radiohead (approved):

PR #162, opened 2023-11-22
[codegen] Consistent Generated Model Naming

IfSentient:
Use sanitized exported kind name as go type for models, instead of using exported machine name. This is consistent with the previous thema codegen for model types.


vlad-diachenko (approved):
lgtm ;)

radiohead (approved):

ferruvich (approved):

PR #181, opened 2024-01-16
Expose subrouter's current path

julienduchesne:
To be used when creating an OpenAPI spec programatically 
As we're working with subrouters, we need to keep track of the current subpath 
By exposing it, we don't have to duplicate that information


IfSentient (approved):
lgtm

PR #182, opened 2024-01-16
[WIP][Proposal] Changes to Schema/Object interfaces

IfSentient:
This WIP/Proposal PR is to make changes to the current `resource.Schema`/`resource.Kind` interfaces to remove some of the abstractions around kubernetes, and explicitly bring forward kubernetes compliance in the interfaces.

The changes here are to make working with kubernetes' SDK's simpler with the types used by the app-sdk, but to preserve the convenience aspects of the app-sdk. The following are the proposed changes:

* `resource.Schema` -> `resource.Kind`. 
    * Right now, there is little difference between a Schema and what kubernetes calls a kind. Since a kind within kubernetes has a group, version, and kind, identically to a schema, we move to use the terminology of kubernetes for this.
    * We also will now hoist the marshal/unmarshal out of the Object implementation and into the Kind, as the Kind should have the knowledge to marshal/unmarshal instances of itself.
    * *UNDECIDED*: do we also want to represent sets of kinds within the same GroupVersion (apiVersion), or the same Kind for multiple versions? I would love discussion/feedback/ideas on this.
* `resource.Object` replacement (right now called `resource.Object2` for simplicity)
    * Explicitly make compliance with kubernetes' `runtime.Object` and `metav1.Object` part of the interface. This means that the object can now be easily used with most kubernetes' SDKs which accept `runtime.Object` (and expect `metav1.Object` compliance).
    * Rename `SpecObject()` and `Subresources()` to `GetSpec()` and `GetSubresources()` to be a bit more in-line with how the `metav1.Object` interface names its methods.
    * I've kept the `StaticMetadata` concept at the moment, but I'm all ears on opinions for that, and whether to bring in other methods from `resource.Object`.
* Generated Object code is changed to comply with the above changes. The changes to users of the generated code are currently:
    * Custom metadata is no longer accessible via the `Metadata` field in `Object`. Instead, Getters/Setters are attached to the object for each custom field in `metadata`, which get/set the value in `Object.ObjectMeta.Annotations`.
    * The `Metadata` field is now called `ObjectMeta` (this has to be done to compose `metav1.ObjectMeta`'s functions onto the `Object`. We could get around this by generating a _lot_ more code, if keeping the `Metadata` key is preferred).

I have not made/tested codegen changes for thema, though now the generated `Kind` implementation will have to vary based on whether thema is being used or not (for non-thema, the kind can just use ordinary JSON unmarshal).

There are TODOs in-code and docs that are invitations for further discussion.

Naming is most definitely not final, I'd love input on the names of things.


charandas (commented):
Went through the changeset.

* Overall, love the refactor and intent of this change.
* ObjectMeta as the field name instead of Metadata works! I like less generated code.
* Object2 - not sure about a number in the name of a type. Did you just mean this as temporary?
* UntypedKind and TypedKind: the naming is golden.
* If StaticMetadata is gonna continue existing, I vote for convinence methods on it similar to what the top level has.

resource/kind.go
```diff
@@ -0,0 +1,319 @@
...
+	GetSpec() any
+	GetSubresources() map[string]any
+	GetStaticMetadata() StaticMetadata
+	SetStaticMetadata(metadata StaticMetadata)
+}
+
+// MultiVersionKind is a collection of Kinds which have the same Group and Kind, but different versions.
+// This may allow us to include conversion as a component of the interface.
+// TODO: not sure if necessary
+type MultiVersionKind interface {
```
ferruvich, line 31: It would be nice to have some examples of use case that make this useful

resource/kind.go
```diff
@@ -0,0 +1,319 @@
...
+	Group() string
+	Kind() string
+	Versions() []Kind
+}
+
+// Group is a collection of Kinds which share a Group and Version, allowing them to be put under the same
+// routing in an API server.
+// TODO: not sure if necessary
+type Group interface {
```
ferruvich, line 40: It would be nice to have some examples of use case that make this useful

resource/kind.go
```diff
@@ -0,0 +1,319 @@
...
+
+// Kind is an interface representing a kubernetes-compatible Kind, which contains information about the Kind's
+// Group, Version, Kind, Plural, and Scope, as well as allowing for creation of a zero-value version of the Kind as an
+// Object (which may contain default values for fields as decided by the implementation).
+// Additionally, Kind contains the logic for reading wire bytes into an Object instance of the Kind,
+// or serializing an Object instance of the Kind into wire bytes.
+// In this sense, Kind combines schema information and a serializer into one interface used to interact with the kind.
+type Kind interface {
+	//Schema <- can't compose because we need Object2
+	// TODO: just update Object instead of using Object2?
```
ferruvich, line 54: What's the main reason behind not updating Object? Is it just because of the breaking changes?

codegen/templates/resourceobject2.tmpl
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+    "github.com/grafana/grafana-app-sdk/resource"
+    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+    "k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type {{.ObjectTypeName}} struct {
+    metav1.TypeMeta `json:",inline"`
+    metav1.ObjectMeta `json:"metadata"`
```
ferruvich, line 15: I'm in favour of having ObjectMeta vs Metadata, better have less generated code 👍🏼

IfSentient:
>     * Object2 - not sure about a number in the name of a type. Did you just mean this as temporary?

Yeah, this is just because I didn't want to erase `Object` as I was working on this initially for compilation/test reasons. Once we think this is the right approach I'll get rid of the old `Object` interface and replace it with what's now `Object2` and make all the necessary changes everywhere to account for it.

resource/kind.go
IfSentient: I intend to replace `Object` once we think this is a good path, initially I was unsure if we'd want to use the same name, and also `Object` is used in a lot of places and I wanted to avoid having to deal with everywhere it was used initially--the intent is to replace `Object` before this gets merged, unless we can think of some better naming.

resource/kind.go
joeblubaugh: I don't know if you need this interface, but it would be nice to make it simple to work with each Kind at multiple versions. This would make writing a migration tool using the app-sdk more straightforward: as long as you have definitions of both versions available to you, you can use those generated types plus k8s APIs to write migration functions.

resource/kind.go
joeblubaugh: Maybe the idea is something like (obviously this is not correct code):
```
groupV1 := &Group{ "slo.api.grafana.com", "v1", [SLOv1{}] }
groupV2 := &Group{ "slo.api.grafana.com", "v2", [SLOv2{}, UserPreferences{}]
router := &VersionedResourceRouter{}
router.HandleResourceGroup(groupV1)
router.HandleResourceGroup(groupV2)
```
and that sets up standard handlers for every type in a group. Don't know where the handler functions come from in that case, though.

resource/kind.go
radiohead: I'm not entirely sure about the use-cases, but I agree that this might be useful for version conversion webhooks.

resource/kind.go
```diff
@@ -0,0 +1,319 @@
...
+	SetStaticMetadata(metadata StaticMetadata)
+}
+
+// MultiVersionKind is a collection of Kinds which have the same Group and Kind, but different versions.
+// This may allow us to include conversion as a component of the interface.
+// TODO: not sure if necessary
+type MultiVersionKind interface {
+	Group() string
+	Kind() string
+	Versions() []Kind
```
radiohead, line 34: NIT:
```suggestion
	Versions() map[string]Kind
```

resource/kind.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// Group is a collection of Kinds which share a Group and Version, allowing them to be put under the same
+// routing in an API server.
+// TODO: not sure if necessary
+type Group interface {
+	Group() string
+	Version() string
+	Kinds() []Kind
```
radiohead, line 43: Hm, shouldn't this be `[]MultiVersionKind`? Or maybe even `map[string]MultiVersionKind`?

IfSentient:
This PR has now evolved into the first of a few PRs into a new branch to refactor Object/Kind logic here: https://github.com/grafana/grafana-app-sdk/pull/200

IfSentient:
This PR has been split into several PRs which are ready to merge to main from a branch here: https://github.com/grafana/grafana-app-sdk/pull/218

Issue #189, opened 2024-01-26
[Bug] Unable to install grafana-app-sdk@latest


L2D2Grafana:
#### Description
Error
```
go install github.com/grafana/grafana-app-sdk/cmd/grafana-app-sdk@latest

go: github.com/grafana/grafana-app-sdk/cmd/grafana-app-sdk@latest (in github.com/grafana/grafana-app-sdk@v0.14.7):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```
Perhaps it is due to this line https://github.com/grafana/grafana-app-sdk/blob/main/go.mod#L6 

Successful older versions install
```
go install github.com/grafana/grafana-app-sdk/cmd/grafana-app-sdk@v0.9.6
```

IfSentient:
Looks like this is an issue since the pinning of `github.com/getkin/kin-openapi` (for codegen) with the `replace` directive in the `go.mod` file. I think a temporary fix for this should be cloning `kin-openapi` into the `internal` package and updating `deepmap/oapi-codegen` to use the cloned internal one. I'll get a PR up to do this, and we can see if it resolves the issue.

The longer-term fix is to split the CLI and codegen packages into a separate go module from the libraries (as the reason the version has to be pinned with the replace directive is that `grafana-plugin-sdk-go` uses it as well, and that is used by the `plugin` package). This will allow a cleaner set of dependencies for both the CLI and libraries.

IfSentient:
We should also probably add a point to the workflow to test this. In the meantime, a workaround is:
1. Clone the repo
2. Run `make build`
3. Copy `target/grafana-app-sdk` into your `$GOPATH/bin` (default location for `go install`) or another executable path

L2D2Grafana:
Also ran into this on `make build`
```
› mage
# github.com/grafana/grafana-plugin-sdk-go/experimental/e2e/storage
../vendor/github.com/grafana/grafana-plugin-sdk-go/experimental/e2e/storage/openapi.go:142:20: cannot range over op.Responses (variable of type *openapi3.Responses)
Error: error compiling magefiles
```

radiohead:
@L2D2Grafana @IfSentient you can download the binary from GH releases page - it's how we got it automated in a few places.

radiohead:
Here's a script we're using together with a Makefile for installing the SDK:
```sh
#!/bin/sh

set -e

version=$1
[...]

PR #190, opened 2024-01-29
Separate into modules for plugin and CLI to avoid versioning issues

IfSentient:
This PR introduces extra go submodules for the `cmd/grafana-app-sdk` and `plugin` directories, and a `go.work` to allow them to incorporate properly. It also moves the `codegen`, `kindsys`, and `internal` packages into the `cmd/grafana-app-sdk` submodule. This accomplishes two major things:
1. The CLI and library dependencies are no longer "mixed." The CLI still have some upward dependencies on the`resource` and `k8s` packages, but now using any of the libraries (`resource`, `k8s`, `operator`, etc.) does not require the dependencies of the `codegen` package or CLI. The `plugin` package has been similarly isolated because dependency versions in the root module will overwrite the ones in the submodule, and the `codegen` package requires an older version of `getkin/kin-openapi` than `plugin` requires.
2. Removes the need for the `replace` directive version pinning, which was preventing `go install` from working correctly, and _also_ required that a user of the app-sdk as a library also provided that `replace` directive in their `go.mod`.

Resolves https://github.com/grafana/grafana-app-sdk/issues/189


radiohead (approved):

Makefile
```diff
@@ -29,6 +29,8 @@ $(LINTER_BINARY):
 .PHONY: test
 test:
 	go test -count=1 -cover -covermode=atomic -coverprofile=$(COVOUT) ./...
+	go test -count=1 -cover -covermode=atomic -coverprofile=$(COVOUT) ./plugin/...
+	go test -count=1 -cover -covermode=atomic -coverprofile=$(COVOUT) ./cmd/grafana-app-sdk/...
```
radiohead, line 33: Good catch!

.github/workflows/release.yml
```diff
@@ -25,6 +25,12 @@ jobs:
     # make test
     - name: Test
       run: go test -v ./...
+    # make test
+    - name: Test CLI module
+      run: go test -v ./cmd/grafana-app-sdk/...
+    # make test
+    - name: Test plugin module
+      run: go test -v ./plugin/...
```
radiohead, line 33: Hm, do you think we can update our workflows to rely on `make` instead? I'm not happy with the duplication we have here.

IfSentient: IIRC there was some problem with using make in the workflow, but let me try and see what happens.

IfSentient: Looks like it worked, nvm, good call

radiohead (approved):

PR #200, opened 2024-02-01
Refactor of the Object interface to be a superset of the kubernetes interfaces, introduce Kinds

IfSentient:
This PR supersedes https://github.com/grafana/grafana-app-sdk/pull/182 and is the first move toward the actual implementation of the ideas present in that PR. Reviewers of https://github.com/grafana/grafana-app-sdk/pull/182 have also been added to this PR. *EDIT: I am unable to add some reviewers to this PR for unexplained GitHub reasons*

This PR is a pull into a branch which is not main, as there will be several PRs to get this into a working (compiling) shape, then that branch will be pulled into main. All the work has been done to get this into a working shape already, but the PR for that would be far too large, so it is being split into these component PRs, to allow for reviews and comments of each part of the design/implementation process. This PR is _just_ the `resource` package, which introduces:
* Changes to the `Object` interface to make it a superset of the kubernetes `runtime.Object`, `metav1.Object`, and `schema.ObjectKind`, along with additional methods.
* Changes to the `ListObject` interface to also make it a superset of `runtime.Object`, `schema.ObjectKind`, and `metav1.ListInterface`
* Removal of the Simple and Store objects in favor of a few basic `Object` implementations: `UntypedObject`, which contains unstructured spec and subresource data; `TypedSpecObject`, which contains typed spec data, `TypedSpecStatusObject`, which contains typed spec and status data; and `TypedObject`, which contains typed spec data and a typed catalog of subresources.
* Since the `Object` interface no longer has the `Unmarshal` method, introduce a `Codec` interface for reading/writing `Object` interfaces, and a `Kind` struct type which unifies a `Schema` and one or more `Codec`s based on the kind encoding. Anything which needs to read/write bytes (`ClientRegistry`, and things which use it) now accept the `Kind` as a parameter instead of `Schema`.
* ~Deprecation of `Schema` and its associated implementations and groups in favor of `Kind`, which is `Schema` + an encoder and decoder for the object.~
* ~Simple implementations of `Kind`: `TypedKind` and `UntypedKind`~
* Updates to the various stores to deal with these changes

There are a few open questions for reviewers noted in `TODO` comments, namely around whether we need certain methods, or if certain method signatures should be expanded upon. Other feedback is more than welcome.


IfSentient:
@charandas @joeblubaugh I was unable to add you two to this PR as reviewers, so I'm notifying you here.

resource/kind.go
```diff
@@ -0,0 +1,431 @@
...
+type UntypedKind struct {
+	// GVK is the group, version, and kind of the UntypedKind, returned by Group(), Version(), and Kind() respectively
+	GVK schema.GroupVersionKind
+	// PluralKind is the plural name of the Kind, returned via Plural()
+	PluralKind string
+	// KindScope is the scope of the Kind, returned by Scope()
+	KindScope SchemaScope
+	// ZeroObject is the *UntypedObject returned by ZeroValue(). This only needs to be non-nil if you do not want to use
+	// an empty *UntypedObject as the ZeroValue() return.
+	ZeroObject *UntypedObject
```
radiohead, line 130: Do we have to make it a pointer? In general we should try to avoid pointers – they are prone to null-pointer errors and they always escape to the heap.

resource/kind.go
```diff
@@ -0,0 +1,431 @@
...
+		OwnerReferences:            obj.GetOwnerReferences(),
+		Finalizers:                 obj.GetFinalizers(),
+		ManagedFields:              obj.GetManagedFields(),
+	}
+	m["spec"] = obj.GetSpec()
+	for k, v := range obj.GetSubresources() {
+		m[k] = v
+	}
+	return json.NewEncoder(out).Encode(m)
+}
```
radiohead, line 208: I think we should separate data types (e.g. `UntypedKind` / `TypedKind`) from serialisation logic. The latter should be implemented separately.

resource/simplestore.go
```diff
@@ -5,33 +5,23 @@ import (
...
-	CommonMetadata CommonMetadata `json:"objectMetadata"`
-	CustomMetadata CustomMetadata `json:"customMetadata"`
-	Subresources   map[string]any `json:"subresources"`
-}
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
 
 // ObjectMetadataOption is a function which updates an ObjectMetadata
-type ObjectMetadataOption func(o *CommonMetadata)
+type ObjectMetadataOption func(o *metav1.ObjectMeta)
```
radiohead, line 13: Should we create aliases for these types to avoid those imports in end-user code?

resource/typedobject.go
```diff
@@ -0,0 +1,590 @@
...
+type TypedSpecObject[T any] struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              T `json:"spec"`
+	// TODO: should this be map[string]any, or map[string]json.RawMessage?
+	// json.RawMessage is clearer about what it is, but it locks it to being JSON bytes
+	// any doesn't tell you anything, so the user needs to have a priori knowledge or do type checks/reflection
+	// It started as json.RawMessage, but for usefulness in SimpleStore I changed it to any--but SimpleStore
+	// may also be deprecated soon...
+	Subresources map[string]any `json:"-"`
```
radiohead, line 34: I'm not a huge fan of having semi-typed objects. Do you think we can somehow replace this with just the typed Status subresource (since it's the mostly used one) and leave the rest up to specific implementations (e.g. if someone wants other subresources, they can define their own object types by embedding this one)?

resource/kind.go
```diff
@@ -0,0 +1,431 @@
...
+// Additionally, Kind contains the logic for reading wire bytes into an Object instance of the Kind,
+// or serializing an Object instance of the Kind into wire bytes.
+// In this sense, Kind combines schema information and a serializer into one interface used to interact with the kind.
+type Kind interface {
+	KindReadWriter
+	Schema
+}
+
+// NewKindGroup returns a new KindGroup with the provided group and version
+func NewKindGroup(group, version string) *KindGroup {
```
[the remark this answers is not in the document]
IfSentient: True, but kubernetes clients are based on both group and version (the `/apis/myapp.ext.grafana.com/v1/` part), so it made sense to have some level of grouping of kinds under the same group _and_ version.

resource/kind.go
IfSentient: `*UntypedObject` implements `Object`, `UntypedObject` doesn't. The implementations have to be pointer structs due to the `Set` methods for fields.

resource/typedobject.go
IfSentient: Yeah, I was very on the fence on even having the subresource list, it only exists because of `SimpleStore`, though honestly we could get rid of it here and work around it in `SimpleStore` (which I believe we will be deprecating anyway).

resource/kind.go
```diff
@@ -0,0 +1,431 @@
...
+
+// KindReader describes any type capable of reading one or more Kind object bytes (in wire format) into a specific or generic Object2.
+// KindReaders may be specific to a particular group/version/kind, or generic across any or all of group, version, and kind.
+// See TypedKind and UntypedKind as examples of KindReader implementations.
+type KindReader interface {
+	// Read consumes the wire-format bytes contained in the io.Reader, and unmarshals them into an instance of the
+	// Kind as an Object2.
+	// It MAY return an error if the provided bytes are not of an expected group, version, and/or kind,
+	// and MUST return an error if the provided bytes are not of the proper shape to be unmarshaled as a kind, or cannot be unmarshaled for any other reason.
+	// TODO: should this _also_ consume a wire format for identifying the type of bytes on the wire, or always assume JSON?
```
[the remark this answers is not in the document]
IfSentient: The problem here being that the code invoking the reader is typically supplied the reader in advance, because kind reading may be non-standard (not just straight marshal/unmarshal). So the client has bytes of some content encoding that it knows (json, yaml, etc.), and was given a `KindReader` to use on creation, rather than at request time (to make the actual API for a user simpler).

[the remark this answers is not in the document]
radiohead: If that's the case then the only option would be to pass the encoding as an argument, yes.

resource/kind.go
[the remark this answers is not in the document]
radiohead: Considering that all our workflows are kind-centric and not group-centric, I don't see any use-cases for a Group of specific Kind versions.

I don't think the implementation details of Kubernetes clients matter here.

resource/typedobject.go
radiohead: I suggest doing that then 👍

resource/kind.go
IfSentient: So right now the separation is that a `Schema` is information about the data type, and a `Kind` combines that with serialization logic. Perhaps we have both `UntypedSchema`/`TypedSchema` in addition to the `UntypedKind`/`TypedKind`? The combination with serialization logic is mainly so that serialization isn't incumbent upon clients (like the `NegotiatedSerializer` in the k8s client), given that kind implementations may vary in logic (the easiest example right now is legacy thema support). 

Perhaps we also explicitly create typed/untyped `KindReadWriter` implementations, which explicitly say that they use bare-bones marshal/unmarshal, and then call out that `TypedKind`/`UntypedKind` use them?

IfSentient: We could also change nomenclature so that `Schema` -> `Kind`, and `Kind` -> `SerializableKind`, or something to that effect, to make things clearer.

radiohead: > The combination with serialization logic is mainly so that serialization isn't incumbent upon clients (like the NegotiatedSerializer in the k8s client), given that kind implementations may vary in logic (the easiest example right now is legacy thema support).

I'm not sure I follow the line of thought here. Why is forcing codecs on the clients a problem? At the end of the day only the clients have enough information to be able to (de)serialise from / to specific wire formats, no?

IfSentient: I keep going in circles with this in my head. We've gone from (de)serialization being controlled by the `Object` implementation itself, to now externalizing (de)serialization into the container for schema information so it can be better re-used and customized, I keep thinking it might be nice to further externalize it, but we do need to keep the option open for a particular kind (GVK) to (de)serialize differently, I _think_. But I'm also not adverse to what is essentially a `NegotiatedSerializer`-type implementation where _all_ kinds for a client (de)serialize the same way, it would just add a new restriction to the SDK that hasn't existed prior (and kind of throws a wrench in thema support, but I might be able to work around it).

This is _also_ something we could, in theory, change to later, by removing the (de)serialization logic from `Kind`, though it would be a breaking change, it would be no more a breaking change than doing the same thing right now, so we could also punt this conversation to a future time.

radiohead: I'm OK with keeping the codec logic in the Kind. I suggest removing all other fields from structs though and instead delegating them to `Schema`, i.e. something like:

```golang
type Codec interface {
  Encode(io.Writer, Object) error
  Decode(io.Reader, Object) error
}

type Kind struct {
  Schema Schema
  Codecs map[KindEncoding]Codec
}
```

IfSentient: That looks like a good approach, thank you! I think this suggested approach is much clearer than the mess I've gotten myself into.

resource/kind.go
```diff
@@ -22,334 +18,162 @@ const (
...
 
-// Kinds returns the list of kinds registered with the group
-func (k *KindGroup) Kinds() []Kind {
-	return k.kinds
+// Kind is a struct which encapsulates Schema information and Codecs for reading/writing Objects which are instances
+// of the contained Schema. It implements Schema using the Schema field.
+type Kind struct {
+	Schema Schema
+	Codecs map[KindEncoding]Codec
 }
```
radiohead, line 44: NIT: seeing as we're just re-implementing `Schema` methods in the Kind, we might as well just embed it. As much as I try to avoid embedding types I think it's one of the few cases where it makes sense.
```suggestion
type Kind struct {
	Schema
	Codecs map[KindEncoding]Codec
}
```

resource/kind.go
```diff
@@ -22,334 +18,162 @@ const (
...
-			Group:   group,
-			Version: version,
-			Kind:    kind,
-		},
-		PluralKind: fmt.Sprintf("%ss", strings.ToLower(kind)),
-		KindScope:  NamespacedScope,
-		ZeroObject: zeroObject,
+// Codec is a nil-safe way of accessing the Codecs map in the Kind.
+// It will return nil if the map key does not exist, or the key is explicitly set to nil.
+func (k *Kind) Codec(encoding KindEncoding) Codec {
```
radiohead, line 96: Question – do we want to redefine Read / Write methods on the Kind?

Something like:
```go
func (k Kind) Read(src io.Reader, enc KindEncoding) (Object, error) {
  c, ok := k.Codecs[enc]
  if !ok {
    return fmt.Errorf("no available codec for encoding: %s", enc)
  }

  res := k.Schema.ZeroValue()

  if err := c.Read(src, res); err != nil {
    return nil, err
  }

  return res, nil
}
```

resource/kind.go
```diff
@@ -0,0 +1,188 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type KindEncoding string
```
radiohead, line 12:
```suggestion
// KindEncoding is the wire encoding of the Kind objects.
type KindEncoding string
```

resource/kind.go
```diff
@@ -0,0 +1,188 @@
...
+	out.SetGroupVersionKind(schema.FromAPIVersionAndKind(apiVersion, kind))
+
+	// Metadata
+
+	// Spec
+	// Reflection?
+
+	// Subresources
+	// Reflection?
+	*/
```
radiohead, line 117: Do we still need this?

radiohead (commented):
Looks good! Just checking if the commented code is still relevant or not.

radiohead (approved):

PR #211, opened 2024-02-16
[Kind/Object Refactor] Codegen Updates

IfSentient:
This PR updates code generation to comply with the new `resource.Object` interface, and introduce `resource.Kind` and `resource.Codec`. *There are a few substantive changes to the generated code*:
* `ExtraFields` has been removed _entirely_ from generated metadata. It is no longer used, as all kubernetes fields are now exposed directly from the `Object.ObjectMeta` (or associated `Object.GetX`/`Object.SetX` functions)
* `Object.Metadata` has been changed to `Object.ObjectMeta`, and is now a kubernetes `metav1.ObjectMeta` object, containing _only_ kubernetes metadata. _Custom_ metadata fields are now stored only in Annotations, but can be retrieved/set with `GetX`/`SetX` generated methods on `Object`.
* A new file named `x_codec_gen.go` is now generated, which has the marshal/unmarshal logic for JSON. This file will be different if you are using CUE- vs Thema-format kinds. Parsing thema-format kinds still uses lineage binding, but some of the logic that formerly lived in grafana-app-sdk/k8s to reformat metadata is now in the generated codec. This should be non-impacting to usage, but something to note (the unmarshal logic has changed, but should be transparent).

Full codegen changes are visible form the diffs on the `golden_generated` files.


codegen/jennies/typescript.go
```diff
@@ -42,15 +128,16 @@ func (j TypeScriptTypes) Generate(kind codegen.Kind) (codejen.Files, error) {
...
+		/*bytes, err := generateTypescriptBytes(ver.Schema, kind.Name(), cfg)
 		if err != nil {
 			return nil, fmt.Errorf("error generating TypeScript for kind '%s', version '%s': %w", kind.Name(), ver.Version, err)
 		}
 		return codejen.Files{codejen.File{
 			RelativePath: fmt.Sprintf("%s_types.gen.ts", kind.Properties().MachineName),
 			Data:         bytes,
 			From:         []codejen.NamedJenny{j},
-		}}, nil
+		}}, nil*/
```
radiohead, line 139: Can we drop this code?

codegen/jennies/typescript.go
```diff
@@ -62,26 +149,80 @@ func (j TypeScriptTypes) Generate(kind codegen.Kind) (codejen.Files, error) {
...
+		/*bytes, err := generateTypescriptBytes(v.Schema, kind.Name(), cfg)
 		if err != nil {
 			return nil, fmt.Errorf("error generating TypeScript for kind '%s', version '%s': %w", kind.Name(), v.Version, err)
 		}
 		files = append(files, codejen.File{
 			RelativePath: fmt.Sprintf("%s/%s/types.gen.ts", kind.Properties().MachineName, v.Version),
 			Data:         bytes,
 			From:         []codejen.NamedJenny{j},
-		})
+		})*/
```
radiohead, line 160: Same – is this intended to be left over?

radiohead (approved):
I _think_ everything looks good. Just a couple of left-over commented-out code.

codegen/jennies/typescript.go
IfSentient: Yup, that's an accidental left-over

radiohead (approved):

PR #221, opened 2024-02-28
Optimise allocations when LISTing objects

radiohead:
### WHAT

This commit attempts to optimise some superfluous allocations which happen when LISTing objects by using pre-allocations and in-place slice appending instead of allocating multiple layers of slices everywhere.

It also ensures that `listObjectWrapper` used by informers correctly handles ListMetadata.

### WHY

We've noticed that with a large amount of objects in the API server, listing those allocates much larger amount of memory compared to `kubernetes-sigs/controller-runtime`. This is an attempt at improving the memory usage by tackling some low-hanging fruits.


IfSentient (approved):

radiohead:
Going to close this one, since we didn't see good enough memory usage improvements and with `v0.15.0` now being a stable release, we can continue our investigations there instead, in case we still have too much overhead.

PR #231, opened 2024-03-14
API Server Poc

toddtreece:


simple/apiserver.go
```diff
@@ -184,8 +183,9 @@ func NewAPIServerOptions(groups []apiserver.ResourceGroup, out, errOut io.Writer
 	}
 
 	o.RecommendedOptions.Admission.Plugins = admission.NewPlugins()
-	for _, g := range groups {
-		for _, r := range g.Resources {
+	for gid, g := range groups {
+		for rid, _ := range g.Resources {
+			r := &groups[gid].Resources[rid]
```
toddtreece, line 188: @IfSentient the issue with multiple versions was the reference in the `range` loop changing

PR #237, opened 2024-03-28
[docs] Add Migration Notes Section

IfSentient:
Add new docs section for migrations, include migration notes for 0.14x -> 0.15.x

### DIFF
--- a/docs/migrations/v0.15.md
+++ b/docs/migrations/v0.15.md
@@ -2,7 +2,9 @@
 
 [Release notes](https://github.com/grafana/grafana-app-sdk/releases/tag/v0.15.0)
 
-## What Changed
+*Upgrading from*: v0.14.x
+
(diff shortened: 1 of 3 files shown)


docs/migrations/v0.15.md
```diff
@@ -0,0 +1,28 @@
+# v0.15
```
ferruvich, line 1: Is this document intended to be explaining a migration from a version **before** v0.15.x to v0.15.x? 
Asking this as it might generate some confusion on developers which are in versions before v0.14.

IfSentient: Good point, I think you can migrate from a few versions back without any major impact, but I'll call out that this doc is for going from v0.14.x to v0.15

radiohead (approved):

ferruvich (approved):
great work!

PR #241, opened 2024-04-04
Updating Operator docs

ferruvich:
# What

This PR updates `docs/operators.md` file, adding new changes related to how to build operators using `grafana-app-sdk`.

# Why

So our documentation for operators is aligned with the current SDK status

### DIFF
--- a/docs/operators.md
+++ b/docs/operators.md
@@ -25,7 +25,9 @@ This watcher adds some internal finalizer logic to make sure events cannot be mi
 and adds a fourth hook: `Sync`, which is called when a resource _may_ have been changed during operator downtime, 
 but there isn't a way to be sure (with a vanilla Watcher in a kubernetes-like environment, these events would be called as `Add`).
 
-<!-- TODO brifely describe the retry logic for this -->
+A Reconciler has its reconciling logic described under the `Reconcile` function.
+The `Reconcile` flow allows for explicit failure (returning an error), which uses the normal retry policy of the `operator.InformerController`, or supplying a `RetryAfter` time in response explicitly telling the `operator.InformerController` to try this exact same Reconcile action again after the request interval has passed.
+As for the watcher, the SDK also offers an _Opinionated_ reconciler, designed for kubernetes-like storage layers, called `operator.OpinionatedReconciler`, and adds some internal finalizer logic to make sure events cannot be missed during operator downtime.
 
 ## Event-Based Design
 
--- a/docs/operators.md
+++ b/docs/operators.md
@@ -14,7 +14,7 @@ An operator consists, broadly, of collections of runnable controllers, one type 
 but a user can easily extend this by having a new controller which implements the `operator.Controller` interface.
 
 The controller offered by the SDK is the `operator.InformerController`, which is a controller that is composed of three sets of objects:
-* **Informers**, which are given a particular CRD and will notify the controller on changes;
+* **Informers**, which are given a particular CRD and will notify the controller on changes - when resources change, Watchers and Reconcilers will be triggered, performing the according actions;
 * **Watchers**, which subscribe to changes for a particular CRD kind and will be notified about any changes from a relevant Informer. Multiple Watchers can watch the same resource kind, and when a change occurs, they will be called in the order they were added to the controller.;
 * **Reconcilers**, which subscribe to changes in the state of a particular CRD kind and will be noticied about any changes from a relevant Informer, its objective is to ensure that the current state of resources matches the desired state. Multiple Reconcilers can watch the same resource kind, and when a change occurs, they will be called in the order they were added to the controller.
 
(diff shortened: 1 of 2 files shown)


docs/operators.md
```diff
@@ -36,7 +40,7 @@ While the SDK is geared toward this type of design, you can still put all of you
 if you either need to have calls be completely synchronous, or the business logic is very simple.
 
 **Forthcoming:** the SDK codegen utility offers the ability to create a simple CRUDL backend plugin API 
-and an operator template given only one or more Schemas defined in CUE (see [Schema Management](schema_management.md)). 
+and an operator template given only one or more Schemas defined in CUE (see [Kinds documentation](custom-kinds/README.md)).
```
ferruvich, line 43: This is pointing to docs created in https://github.com/grafana/grafana-app-sdk/pull/240/files

IfSentient: This is actually no longer forthcoming, `grafana-app-sdk project component add backend` does this, though the CRUDL plugin backend API is also something we want to slowly move away from in favor of direct API server access.

docs/operators.md
```diff
@@ -14,9 +14,9 @@ An operator consists, broadly, of collections of runnable controllers, one type
...
-* **Informers**, which are given a particular CRD and will notify the controller on changes, and
-* **Watchers**, which subscribe to changes for a particular CRD kind and will be notified about any changes from a relevant Informer
-Multiple Watchers can watch the same resource kind, and when a change occurs, they will be called in the order they were added to the controller.
+* **Informers**, which are given a particular CRD and will notify the controller on changes;
+* **Watchers**, which subscribe to changes for a particular CRD kind and will be notified about any changes from a relevant Informer. Multiple Watchers can watch the same resource kind, and when a change occurs, they will be called in the order they were added to the controller.;
+* **Reconcilers**, which subscribe to changes in the state of a particular CRD kind and will be noticied about any changes from a relevant Informer, its objective is to ensure that the current state of resources matches the desired state. Multiple Reconcilers can watch the same resource kind, and when a change occurs, they will be called in the order they were added to the controller.
```
radiohead, line 19: We probably should explain the difference between watchers and reconcilers and that it's enough to specify one or the other. As it stand right now one may erroneously think that it's necessary to define all three – informer, watcher and a reconciler.

docs/operators.md
```diff
@@ -14,9 +14,9 @@ An operator consists, broadly, of collections of runnable controllers, one type
 but a user can easily extend this by having a new controller which implements the `operator.Controller` interface.
 
 The controller offered by the SDK is the `operator.InformerController`, which is a controller that is composed of two sets of objects:
```
radiohead, line 16: This says "two sets" but we're enumerating three different ones now.

docs/operators.md
```diff
@@ -58,37 +63,38 @@ func main() {
...
-	// We'll use an opinionated watcher, as that's best practice if you don't need to do anything fancy
-	watcher, err := operator.NewOpinionatedWatcher(kubeConfig, MyTypeCustomResource)
-	if err != nil {
-		// Do something with the error
-		panic(err)
-	}
-	watcher.AddFunc = func(ctx context.Context, object runtime.Object) error {
-		fmt.Println("Hey, a resource got added!")
-		return nil
+	watcher := simple.Watcher{
```
radiohead, line 87: @IfSentient should we recommend using Reconcilers instead of Watchers by default, or do you think it doesn't matter?

IfSentient: I think at the moment it doesn't matter--kubernetes uses watchers for everything, so that might be a more familiar paradigm for some. Depending on the direction we take informers, reconcilers may be a better UX in the future (for example, if we have an informer that doesn't contain a cache, then you won't have a previous version of an object in a watcher's `Update` method), but at the moment both watchers and reconcilers are using the exact same [AddEventHandler](https://github.com/grafana/grafana-app-sdk/blob/main/operator/informer_kubernetes.go#L121) logic in the kubernetes SharedIndexInformer.

radiohead: For context, I was basing my assumption on the fact that the recommended way of writing controllers in "vanilla" k8s is https://github.com/kubernetes-sigs/controller-runtime, which only uses the reconciler pattern.

docs/operators.md
ferruvich: ugh missed that, fixed!

docs/operators.md
ferruvich: Mmh it might make sense to use reconciler pattern here just to keep stuff as simple as possible? 
We can then add a link to the example folder, and might I suggest having three/four examples there:
- Watcher / Reconciler using `simple` package;
- Opinionated Watcher / Reconciler using `operator` package;

wdyt?

docs/operators.md
ferruvich: Don't you think the descriptions below (line 21 to 30) define well enough the differences between Watchers and Reconcilers? What additional information you think it's worth adding here? 
Totally agree about the need to specify users don't need both watcher and reconcilers to make things work. I added that on  line 32 👍🏼

docs/operators.md
radiohead: SGTM 👍

docs/operators.md
radiohead: Ah, I've missed the descriptions due to the diff being collapsed. I think the way it is right now is good!

docs/operators.md
radiohead: LGTM!

docs/operators.md
ferruvich: I'll use the reconciler pattern here then, and I'll work on the `examples` package in a followup 👍🏼

ferruvich: Done!

radiohead (approved):

PR #245, opened 2024-04-08
Updating operator examples

ferruvich:
# What

Updating operator examples in `examples/operator` folder to be more in line with our idea of how to write operators using the SDK.
This PR adds the following:
- A basic watcher using the `basic` package and the Watcher pattern (in `basic/watcher/main.go`);
- A basic reconciler using the `basic` package and the Reconciler pattern (in `basic/reconciler/main.go`);
- An opinionated watcher using the `operator` package and the Watcher pattern (in `opinionated/watcher/main.go`);
- An opinionated reconciler using the `operator` package and the Reconciler pattern (in `opinionated/reconciler/main.go`).

# Why

So we do have clear and useful examples for our users.


examples/operator/basic-watcher/README.md
```diff
@@ -1,6 +1,7 @@
-# Basic Example
+# Basic Watcher Example
 
-This example code is a super-basic one-file example of a very simple kubernetes custom resource operator using the SDK. It doesn't handle synchronization or missed events, and is extremely noisy due to not filtering updates.
+This example code is a super-basic one-file example of a very simple kubernetes custom resource operator with a Watcher, using the SDK. 
+It doesn't handle synchronization or missed events, and is extremely noisy due to not filtering updates.
```
IfSentient, line 4: This actually isn't true for `simple.Operator.WatchKind`, which wraps the provided `simple.Watcher` in an `operator.OpinionatedWatcher`.

ferruvich: Yeah I still have to review the readmes 😬

examples/operator/basic/README.md
```diff
@@ -1,50 +1,64 @@
...
 The operator will monitor for changes to any BasicCustomResource in your cluster, and log them. To demonstrate, run:
+
 ```shell
 $ kubectl create -f example.yaml
 ```
-You should see a log line for the added resource. The "noisy" part comes next, as you'll see repeated log lines for near-constant updates by kubernetes itself. These are all metadata updates, and can be filtered in a practical setting. For examples of such filtering, see the [opinionated](../opinionated/README.md) or `/example/operator/boilerplate` examples.
+
+You should see a log line for the added resource. The "noisy" part comes next, as you'll see repeated log lines for
+near-constant updates by kubernetes itself. These are all metadata updates, and can be filtered in a practical setting.
+For examples of such filtering, see the [opinionated](../opinionated/README.md) or `/example/operator/boilerplate`
```
IfSentient, line 39: Since `simple` wraps reconcilers and watchers with their opinionated variants, we may want to instead mention that the logic used in `simple` follows the `opinioned` variants, and introduce code that uses an operator _without_ opinionated logic if need be.

Perhaps this also indicates that `simple.Operator` should have some level of preferences on wrapping reconcilers/watchers, but I also don't want `simple` to become too complex.

ferruvich: To better understand what you're suggesting: should move the example using `simple` package in the `opinionated` section, and use a different approach in the `simple` section which is not using anything having the `opinionated` variant?

IfSentient: I think for the moment, it might be enough to point out that using the `simple` package automatically uses the `opinionated` logic, and re-name `basic` to `simple`. We could introduce a `basic` variant later that doesn't use the opinionated logic at all, but has to build an operator without `simple` (or using `simple` with some "don't-use-opinionated" option yet to be introduced).

IfSentient (approved):

Issue #254, opened 2024-04-15
[docs] Improve Resource Stores Documentation


IfSentient:
The [Resource Stores](https://github.com/grafana/grafana-app-sdk/blob/main/docs/resource-stores.md) doc is a bit of a stub which could use improvement.

PR #267, opened 2024-04-26
[docs] Improve Resource Stores Doc

IfSentient:
Improve the Resource Stores doc with updates and examples, and include a note on the deprecation of `SimpleStore`, informing users to open an issue or discussion if there is a use-case for SimpleStore that TypedStore does not cover.

Resolves https://github.com/grafana/grafana-app-sdk/issues/254

### DIFF
--- a/docs/resource-stores.md
+++ b/docs/resource-stores.md
@@ -45,7 +45,7 @@ The most useful aspect of this example is that we can directly work with the ret
 
 ## Store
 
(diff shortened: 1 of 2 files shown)


docs/resource-stores.md
```diff
@@ -1,16 +1,95 @@
...
+    if _, err := store.Update(ctx, item.GetStaticMetadata().Identifier(), item); err != nil {
+        panic(err)
+    }
+}
+```
+The most useful aspect of this example is that we can directly work with the returned `*v1.MyObject` instances, instead of a `resource.Object` that is typically returned by a Client.
+
+## Store
+
+`resource.Store` is a generic store that allows for working with any kind, not just a single one like `TypedStore` or `SimpleStore`. This comes with the downside that arguments are returned types now use `resource.Object` instead of concrete types. If you're only working with a single kind, prefer using `resource.TypedStore` (you may want to consider multiple `resource.TypedStore` instances for multiple kinds as well if you find the workflow simpler, as the reasource overhead isn't that signficant, considering that `resource.Store` still utilizes a `resource.Client` instance per kind under the hood).
```
radiohead, line 48: It's a bit strange to compare with `SimpleStore` here, because it's not been described yet. I suggest that we either remove this comparison or put the `SimpleStore` section before the `Store` section.

radiohead (approved):

Issue #255, opened 2024-04-15
[docs] How-To for Writing an Operator


IfSentient:
Currently, while we have a fair few docs about design and concepts (and the tutorial), we do not have a best-practices/how-to guide for writing an application operator. The tutorial only briefly touches on the operator code generated by the app-sdk, and many apps will need more complex operator design, so we should have a doc that walks through best practices, when to use a watcher vs a reconciler, deciding what version to watch in a multi-version kind, what logic belongs in a watcher/reconciler vs admission, etc.

PR #269, opened 2024-04-29
[docs] Update Operator Docs

IfSentient:
Rather large docs enhancement, update the operator doc and add a new doc on Writing an Operator, which breaks down the `simple.Operator` and how to use more complex behavior with `operator.Operator`. As part of this enhancement, additions were made to the tutorial doc for adding functionality to the watcher and adding admission control to the operator. Some small fixes to local project setup and codegen were made along the way as a product of testing the new tutorial changes.

Resolves https://github.com/grafana/grafana-app-sdk/issues/255

A future PR should be made to introduce multiple versions to the tutorial to go alongside https://github.com/grafana/grafana-app-sdk/blob/main/docs/custom-kinds/managing-multiple-versions.md, but that is out-of-scope for this PR.


radiohead (approved):
Nice! I just had an idea – do you think we should replace our current examples (since most of them are outdated) with the issue tracker code from the tutorial?

IfSentient:
> Nice! I just had an idea – do you think we should replace our current examples (since most of them are outdated) with the issue tracker code from the tutorial?

Possibly, I think we have some examples that use things the tutorial code doesn't (like more advanced store usage, and a non-simple operator), but there's aspects we can re-use.

PR #277, opened 2024-05-08
[CLI] New codegen updates for API server compatibility

IfSentient:
Ported codegen changes from the `apiserver-poc` branch and updated the `component add` codegen to be compatible. Codegen changes are behind new flags introduced to the CLI:
* `--kindgrouping=kind|group` - `kind` is the existing, default behavior, `group` puts all the kinds for the same group in the same package, and prefixes all the types and functions for a kind with the kind name (excepting the kind `resource.Object` implementation, which retains the same name).
* `--postprocess=true|false` - defaults to `false`, and will be `true` if present (like `--overwrite`). This instructs the CLI to run code generation based on the generated files (right now, just the kubernetes openAPI generation). This is not on by default as it will cause the `generate` command to fail if `go.mod` doesn't list all the dependencies in the generated code.

**Changes not behind flags:**
* A `resource.ListObject` implementation is now generated as part of the codegen. It is not currently used anywhere aside from OpenAPI generation in `--postprocess`, but a separate PR will introduce a default/zero-value `ListObject` to the `Schema` interface, and codegen will be updated to add this implementation to its generated schema. 
* `// +k8s:openapi-gen=true` comments have been added to generated types. These are used by `--postprocess`, but should not impact anything otherwise.


toddtreece (approved):

codegen/jennies/dstutil.go
```diff
@@ -10,6 +10,23 @@ import (
...
+		switch c.Node().(type) {
+		case *dst.TypeSpec:
+			c.Parent().Decorations().Start.Append("// +k8s:openapi-gen=true")
+		case *dst.StructType:
+			// TODO: is there a way of getting to the typedef from the struct cursor?
+			/* c.Parent()
+			//c.Parent().(*dst.TypeSpec).Decs.Start = []string{"// +k8s:openapi-gen=true"}
+			fmt.Println(c.Parent().Decorations())
+			fmt.Printf("%#v\n", c.Parent())
+			x.Decs.Start = append(x.Decs.Start, "// +k8s:openapi-gen=true") */
```
radiohead, line 24: Just for my understanding, is this going to be a blocker or a nice-to-have? IIUC `TypeSpec` should cover all types – did I get it wrong?

codegen/jennies/openapi.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/kube-openapi/cmd/openapi-gen/args"
+	"k8s.io/kube-openapi/pkg/generators"
+
+	"github.com/grafana/grafana-app-sdk/codegen"
+)
+
+// OpenAPI jenny uses kube-openapi to build OpenAPI spec.
+// Because k8s.io/gengo doesn't allow you to use in-memory files (as the Overlay in the config is not accessible),
+// this jenny must be run _after_ the go type has been written to disk.
+type OpenAPI struct {
```
radiohead, line 23: NIT – how much effort would it be for use to just generate k8s-specific OAPI registration code ourselves, without resorting to using `k8s.io/gengo`?

I'm asking because we already have OAPI on our hands during CUE -> Go codegen (since we technically do CUE -> OAPI -> Go) and it would make sense to generate k8s OAPI files as part of normal Go codegen.

That'd let us get rid of the awkward post-processing step.

That said, I think it's totally fine to use post-processing approach for now and keep it that way if it's too much work to get off of it.

radiohead (approved):

codegen/jennies/dstutil.go
IfSentient: Nice-to-have--right now it ends up putting the k8s codegen annotation on every type, even when not applicable. Doesn't hut anything, but makes the code messier.

codegen/jennies/openapi.go
IfSentient: I don't think it would be a impassible hurdle, but it's definitely more work (IIRC the naming of the OpenAPI needs to match the go packaging, and there were a few other gotchas that made it simper to just use `gengo` for the first pass). Long-term we can definitely explore migrating to doing all the codegen ourselves (while possibly still including the k8s gengo comments for if folks want to use them).

codegen/jennies/dstutil.go
radiohead: Got it. Do you mind leaving mentioning explicitly in a comment?

codegen/jennies/openapi.go
radiohead: Alright, sounds good. I _think_ we may want to add an issue to follow up on this but I'll leave it up to you to decide 👍

radiohead (approved):

PR #290, opened 2024-05-23
Pagination in Store and TypedStore

IfSentient:
Auto-pagination in Store and TypedStore, with additional methods for fetching a specific page without using auto-pagination.

To allow for customizing the perPage limit, an additional parameter was added to the `List` method. It will automatically fetch each page and concatenate into a single response. To fetch a single page at a time, a new method called `ListPage` has been introduced.


resource/store.go
```diff
@@ -245,15 +245,48 @@ func (s *Store) ForceDelete(ctx context.Context, kind string, identifier Identif
 }
 
 // List lists all resources of kind in the provided namespace, with optional label filters.
-func (s *Store) List(ctx context.Context, kind string, namespace string, filters ...string) (ListObject, error) {
+// It will auto-list all pages in the response using the perPage limit for requests.
+// To list a single page of results, use ListPage.
+func (s *Store) List(ctx context.Context, kind string, namespace string, perPage int, filters ...string) (ListObject, error) {
```
radiohead, line 250: NIT – I _think_ having `perPage`, `namespace` and `filters` grouped into some kind of options struct would make for better UX. Incidentally it would also make the UX of listing non-namespaced resources better, since you won't need to pass an empty string to list – simply not specifying the ns in the options would take care of that.

resource/typedstore.go
```diff
@@ -169,18 +169,52 @@ func (t *TypedStore[T]) ForceDelete(ctx context.Context, identifier Identifier)
...
 }
 
-// List lists all resources in the provided namespace, optionally filtered by the provided filters
-func (t *TypedStore[T]) List(ctx context.Context, namespace string, filters ...string) (*TypedList[T], error) {
-	resp := &TypedList[T]{}
-	err := t.client.ListInto(ctx, namespace, ListOptions{
+// List lists all resources in the provided namespace, optionally filtered by the provided filters.
+// It will auto-list all pages in the response using the perPage limit for requests.
+// To list a single page of results, use ListPage.
+func (t *TypedStore[T]) List(ctx context.Context, namespace string, perPage int, filters ...string) (*TypedList[T], error) {
```
radiohead, line 175: NIT – Same note about grouping optional args into an options struct.

resource/typedstore.go
```diff
@@ -169,18 +169,52 @@ func (t *TypedStore[T]) ForceDelete(ctx context.Context, identifier Identifier)
...
+	resp := &TypedList[T]{}
+	err := t.client.ListInto(ctx, namespace, options, resp)
 	if err != nil {
 		return nil, err
 	}
 	return resp, nil
 }
 
+// Client returns the underlying Client for this store.
+func (t *TypedStore[T]) Client() Client {
```
radiohead, line 214: QQ – is this required for something in this PR or just a drive-by change?

radiohead (approved):
LGTM. I _think_ grouping args into options makes for better UX but I'm not 100% sure and I don't want to block on that.

resource/typedstore.go
IfSentient: Drive-by in order to have `ListPage` have the same semantic equivalent as in `Store`.

resource/store.go
IfSentient: Yeah, I went back and forth on this one. I can change it to use an options struct specific to stores.

resource/store.go
```diff
@@ -244,16 +250,50 @@ func (s *Store) ForceDelete(ctx context.Context, kind string, identifier Identif
...
-	return client.List(ctx, namespace, ListOptions{
-		LabelFilters: filters,
+	resp, err := client.List(ctx, options.Namespace, ListOptions{
+		Limit:        options.PerPage,
+		LabelFilters: options.Filters,
 	})
+	if err != nil {
+		return nil, err
+	}
+	if resp.GetContinue() == "" {
```
joeblubaugh, line 269: nit: I think this conditional is redundant - if it's ever empty you'll just skip past the `for` to the end of the function.

radiohead (approved):

resource/store.go
IfSentient: Holdover from an initial pass where a new list was created and everything was added to that. Will remove this check.

radiohead (approved):

ferruvich (approved):
LGTM!

PR #296, opened 2024-05-29
[PoC] Use cog to generate go and typescript models + resources

K-Phoen:
Draft PR meant to try out [cog](https://github.com/grafana/cog) as a potential replacement for Go and Typescript codegen purposes.

ToDo:

* [ ] the `thema.CustomKindParser` provided by `grafana-app-sdk` used with cog's CUE parser struggle to correctly handle references to objects defined "above" the currently parsed root. Ex:
    ```cue
    #InnerObject2: { // ← object defined "above" the current root
        name: string
        details: {
            [string]: _
        }
    }

    spec: { // ← current root
        inner: #InnerObject1
        // ...
    }
    ```
    Cog detects this case correctly, gets the path to `InnerObject2` (something like `customKind.lineage._sortedSchemas[1]._#schema.#InnerObject2`), but resolving this path (with `LookupPath()` fails on hidden fields introduced by thema (`_sortedSchemas` and `_#schema`).
    If thema support isn't meant to stick around long-term, this issue might not be worth solving.

Note: this PR isn't necessarily meant to be finished or merged, it's more a sandbox to play with cog + grafana-app-sdk.


radiohead:
@K-Phoen thank you for experimenting with this! This is something that's been on our radar for a while, but we never had any bandwidth for.

Please let us know if we can help get this ready for a review. One notable change we've made since this PR was opened is that we got rid of Thema support, so this should no longer be an issue.

K-Phoen:
Not having to support Thema schemas should simplify things :)

I'll get back to working on this and let you know if I need a hand!

PR #302, opened 2024-06-04
Fix backend plugin codegen

radiohead:
### WHAT

This commit fixes codegen for backend plugin router by ensuring that router service interface uses consistent interface method signatures.

### WHY

To make sure that the codegen works again and produces valid Go code.


radiohead:
This was introduced in https://github.com/grafana/grafana-app-sdk/pull/290 – I unfortunately have missed that. Working with Go templates is hard.

onprem (approved):

ferruvich (approved):

PR #321, opened 2024-06-14
Have NewOpinionatedWatcher and NewOpinionatedReconciler error when finalizer exceeds max length

IfSentient:
Have NewOpinionatedWatcher and NewOpinionatedReconciler error on a finalizer which exceed the kubernetes finalizer max length. Allow a user to customize the finalizer generation function of simple.Operator.

### DIFF
--- a/operator/opinionatedwatcher.go
+++ b/operator/opinionatedwatcher.go
@@ -68,7 +68,7 @@ func NewOpinionatedWatcherWithFinalizer(sch resource.Schema, client PatchClient,
 	}
 	finalizer := supplier(sch)
 	if len(finalizer) > 63 {
-		return nil, fmt.Errorf("finalizer length cannot exceed 63 chars")
+		return nil, fmt.Errorf("finalizer length cannot exceed 63 chars: %s", finalizer)
 	}
 	return &OpinionatedWatcher{
 		client:    client,
--- a/operator/opinionatedwatcher.go
+++ b/operator/opinionatedwatcher.go
@@ -66,10 +66,14 @@ func NewOpinionatedWatcherWithFinalizer(sch resource.Schema, client PatchClient,
 	if client == nil {
 		return nil, fmt.Errorf("client cannot be nil")
 	}
+	finalizer := supplier(sch)
+	if len(finalizer) > 63 {
+		return nil, fmt.Errorf("finalizer length cannot exceed 63 chars")
+	}
 	return &OpinionatedWatcher{
 		client:    client,
 		schema:    sch,
-		finalizer: supplier(sch),
+		finalizer: finalizer,
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


operator/opinionatedwatcher.go
```diff
@@ -66,10 +66,14 @@ func NewOpinionatedWatcherWithFinalizer(sch resource.Schema, client PatchClient,
 	if client == nil {
 		return nil, fmt.Errorf("client cannot be nil")
 	}
+	finalizer := supplier(sch)
+	if len(finalizer) > 63 {
+		return nil, fmt.Errorf("finalizer length cannot exceed 63 chars")
```
jeschkies, line 71:
```suggestion
		return nil, fmt.Errorf("finalizer length cannot exceed 63 chars: %s", finalizer)
```

jeschkies (commented):
Thanks for the fast fix :slightly_smiling_face:

simple/operator.go
```diff
@@ -195,6 +202,9 @@ func (o *Operator) WatchKind(kind resource.Kind, watcher SyncWatcher, options Li
 		return err
 	}
 	ow, err := operator.NewOpinionatedWatcherWithFinalizer(kind, client, func(sch resource.Schema) string {
+		if o.FinalizerGenerator != nil {
+			return o.FinalizerGenerator(sch)
+		}
 		if o.Name != "" {
 			return fmt.Sprintf("%s-%s-finalizer", o.Name, kind.Plural())
```
radiohead, line 209: I would also trim this to just use the name. The kind is already clear (because that's what we're attaching the finalizer to) and `-finalizer` is clear because it's part of `metadata.finalizers`.
```suggestion
			return o.Name
```

An example from built-in Namespace finalizer:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: reloader
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

simple/operator.go
```diff
@@ -226,14 +236,16 @@ func (o *Operator) ReconcileKind(kind resource.Kind, reconciler operator.Reconci
 		return err
 	}
 	finalizer := fmt.Sprintf("%s-finalizer", kind.Plural())
-	if o.Name != "" {
+	if o.FinalizerGenerator != nil {
+		finalizer = o.FinalizerGenerator(kind)
+	} else if o.Name != "" {
 		finalizer = fmt.Sprintf("%s-%s-finalizer", o.Name, kind.Plural())
```
radiohead, line 242: Same as the above – default finalizer name can be trimmed.

radiohead (approved):
LGTM but I think we should also trim default finalizer names.

simple/operator.go
IfSentient: I don't want to change this at the moment without first confirming what existing apps are using--otherwise it may change their finalizers and break operator function. But agreed that we can shorten the defaults once we get the ability to customize the finalizer name into `simple.Operator`.

Issue #301, opened 2024-06-04
ServerResponseError does not cast the kubernetes StatusError type


joeblubaugh:
Kubernetes apimachinery has a "StatusError" type.

The app-sdk has an ServerResponseError type that abstracts certain k8s errors, and it uses an interface to match kubernetes api errors and status fields:
```go
type APIServerResponseError interface {
	error
	StatusCode() int
}
```

However, StatusError has no such method, and instead uses a field to expose the status code:
```go
type StatusError struct {
	ErrStatus metav1.Status
}

...
// metav1.Status
// Status is a return value for calls that don't return other objects.
type Status struct {
	TypeMeta `json:",inline"`
	// Standard list metadata.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
	// +optional
	ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`

	// Status of the operation.
	// One of: "Success" or "Failure".
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
	// +optional
	Status string `json:"status,omitempty" protobuf:"bytes,2,opt,name=status"`
	// A human-readable description of the status of this operation.
	// +optional
	Message string `json:"message,omitempty" protobuf:"bytes,3,opt,name=message"`
	// A machine-readable description of why this operation is in the
	// "Failure" status. If this value is empty there
	// is no information available. A Reason clarifies an HTTP status
	// code but does not override it.
	// +optional
	Reason StatusReason `json:"reason,omitempty" protobuf:"bytes,4,opt,name=reason,casttype=StatusReason"`
	// Extended data associated with the reason.  Each reason may define its
	// own extended details. This field is optional and the data returned
	// is not guaranteed to conform to any schema except that defined by
	// the reason type.
	// +optional
	// +listType=atomic
	Details *StatusDetails `json:"details,omitempty" protobuf:"bytes,5,opt,name=details"`
	// Suggested HTTP return code for this status, 0 if not set.
	// +optional
	Code int32 `json:"code,omitempty" protobuf:"varint,6,opt,name=code"`
}
```

I've encountered circumstances where TypedStore returns a `StatusError` instead of a `ServerResponseError` because this casting code doesn't understand that type:
```go
func (t *TypedStore[T]) Upsert(ctx context.Context, identifier Identifier, obj T) (T, error) {
[...]

PR #349, opened 2024-07-29
[k8s] Properly parse the kubernetes' errors.StatusError type

IfSentient:
Have the k8s package properly parse the kubernetes' errors.StatusError type if returned and return a `*k8s.ServerResponseError` wrapping it. Add tests for error parsing. Resolves https://github.com/grafana/grafana-app-sdk/issues/301

### DIFF
--- a/k8s/gvclient.go
+++ b/k8s/gvclient.go
@@ -15,7 +15,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
-	errors2 "k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


k8s/gvclient.go
```diff
@@ -14,6 +15,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
```
ferruvich, line 18: nit: wdyt about renaming this to k8serrors? I think `errors2` doesn't give enough context when someone notices its usage in this file

ferruvich (approved):
LGTM, left one nit comment

k8s/gvclient.go
IfSentient: Yeah, probably a good idea, that was goland's auto-choice

ferruvich (approved):

PR #378, opened 2024-08-28
Add fieldSelectors everwhere there are label selectors

bcotton:
I believe that watch also supports fieldSelectors.
I may have missed something


operator/informer_customcache.go
```diff
@@ -193,7 +193,7 @@ func (c *CustomCacheInformer) errorHandler(ctx context.Context, err error) {
 
 // NewListerWatcher returns a cache.ListerWatcher for the provided resource.Schema that uses the given ListWatchClient.
 // The List and Watch requests will always use the provided namespace and labelFilters.
-func NewListerWatcher(client ListWatchClient, sch resource.Schema, namespace string, labelFilters ...string) cache.ListerWatcher {
+func NewListerWatcher(client ListWatchClient, sch resource.Schema, namespace string, labelFilters []string, fieldSelectors []string) cache.ListerWatcher {
```
IfSentient, line 196: I think it might be better for us to define some type like 
```go
type ListWatchOptions struct {
    LabelSelectors []string
    FieldSelectors []string
}
```
And consume that, rather than adding to the argument list. We could also use this in the New methods like `NewKubernetesBasedInformerWithFilters` and `NewMemcachedInformerWithLabelFilters`

radiohead: Agree, I would also move the namespace to the options as well. IIRC this is how it's expressed in k8s stdlib too.

bcotton: Just saw your comment @radiohead,  I'll add that also

bcotton: Now we have `simple.ListWatchOptions` and `operator.ListWatchFilterOptions`  that are duplicates. 

Thoughts?

radiohead: @bcotton I would remove `simple.ListWatchOptions` OR make it a type alias. @IfSentient WDYT?

bcotton: Moved `simple.ListWatchOptions` to `operator.ListWatchOptions` and removed `operator. ListWatchFilterOptions`

operator/informer_customcache.go
```diff
@@ -48,19 +48,19 @@ type CustomCacheInformer struct {
...
 // This is analogous to calling NewCustomCacheInformer with a MemcachedStore as the store.
 func NewMemcachedInformer(kind resource.Kind, client ListWatchClient, namespace string, addrs ...string) (*CustomCacheInformer, error) {
-	return NewMemcachedInformerWithLabelFilters(kind, client, namespace, nil, addrs...)
+	return NewMemcachedInformerWithLabelFilters(kind, client, ListWatchOptions{Namespace: namespace}, addrs...)
 }
 
 // NewMemcachedInformerWithLabelFilters creates a new CustomCacheInformer which uses memcached as its custom cache.
 // This is analogous to calling NewCustomCacheInformer with a MemcachedStore as the store.
-func NewMemcachedInformerWithLabelFilters(kind resource.Kind, client ListWatchClient, namespace string, labelFilters []string, addrs ...string) (*CustomCacheInformer, error) {
+func NewMemcachedInformerWithLabelFilters(kind resource.Kind, client ListWatchClient, filterOptions ListWatchOptions, addrs ...string) (*CustomCacheInformer, error) {
```
IfSentient, line 56: I think this should be renamed to `NewMemcachedInformerWithFilters`, as `WithLabelFilters` is not longer correctly descriptive.

IfSentient (changes requested):
One small change, otherwise LGTM

operator/informer_customcache.go
radiohead: Updated.

IfSentient (approved):

Issue #383, opened 2024-08-31
Add preflight checks to fail before buildtime


unlikelyzero:
When attempting to build the project with `make all` I see the following error:

```sh
grafana-app-sdk git:(main) make all
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b target v1.60.3
golangci/golangci-lint info checking GitHub for tag 'v1.60.3'
golangci/golangci-lint info found version: 1.60.3 for v1.60.3/darwin/amd64
golangci/golangci-lint info installed target/golangci-lint
target/golangci-lint-1.60.3 run  ./... ./plugin/...
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`. 
WARN [config_reader] The configuration option `linters.gomnd.settings` is deprecated. Please use the options `linters.gomnd.checks`,`linters.gomnd.ignored-numbers`,`linters.gomnd.ignored-files`,`linters.gomnd.ignored-functions`. 
WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar. 
ERRO Running error: context loading failed: failed to load packages: failed to load packages: failed to load with go/packages: err: exit status 1: stderr: reading go.work: /grafana-app-sdk/go.work:1: invalid go version '1.23.0': must match format 1.23 
make: *** [lint] Error 3
```

At first, I thought the problem was that go could not handle a z stream reference inside of a `go.mod` file. That's not the case.

It seems like go has a general issue with forward versioning as seen here: https://github.com/golang/go/discussions/55092 and the `gvm` project does not look to be healthy. 

Until some of that dust settles, it makes sense to add a preflight check to ensure that this repo fails before buildtime if the environment is wrong.

PR #384, opened 2024-08-31
Add preflight checks for clearer failures and reference go.mod where possible

unlikelyzero:
Closes: https://github.com/grafana/grafana-app-sdk/issues/383

It seems like go has a general issue with forward versioning as seen here: https://github.com/golang/go/discussions/55092 and the `gvm` project does not look to be healthy. 

Until some of that dust settles, it makes sense to add a preflight check to ensure that this repo fails before buildtime if the build environment is wrong.

1. Add a preflight check such that go is detected in the PATH
1. Adds a preflight check such that go version mismatches are more explicit
1. Updates CI to reference go.mod to prevent heartburn and headaches


radiohead (approved):
Thank you for the contribution! I'll give @IfSentient a chance to take a look but otherwise this looks good!

IfSentient (approved):
LGTM, thank you for putting this together!

PR #399, opened 2024-09-12
Codegen: Parse the gvk once

ryantxu:
minor PR mostly to see if I can do it :)

Rather than parsing and creating a GVK for each property, this gets it once and shares the results.


codegen/templates/resourceobject.tmpl
```diff
@@ -66,12 +66,13 @@ func ({{.ObjectShortName}} *{{.TypeName}}) SetSubresource(name string, value any
 }
 
 func ({{.ObjectShortName}} *{{.TypeName}}) GetStaticMetadata() resource.StaticMetadata {
+    gvk := {{.ObjectShortName}}.GroupVersionKind()
```
ryantxu, line 69: seems better to do this once:
https://github.com/kubernetes/kubernetes/blob/v1.31.1/staging/src/k8s.io/apimachinery/pkg/runtime/schema/group_version.go#L300

IfSentient (approved):

Issue #385, opened 2024-09-03
Allow for Runtime and App Separation


IfSentient:
Currently, an app is built targeting a specific runtime--the only available one being an operator with kubernetes webhooks (though custom functionality can be layered on top). To better allow apps to be targeted for different deployment types, such as an [API server](https://github.com/grafana/grafana-app-sdk/issues/151), or running entirely as a multi-tenant (or even single tenant) plugin, or in the future as other runtimes such as WASI, the logic for what an app can do (Control loops, Admission, Conversion, etc.) should be split from how this functionality is called (webhooks for admission/conversion, or inlined in an API server, for example).

The proposal for this is to define an app interface which describes behavior, and runners which can then run the app based on how app interactions from "outside" should be handled (for example, how does a validation request get to the app).

```go
type App interface {
    // App interactions, including, but not limited to:
    Validate(ValidationRequest) (bool, error)
    Mutate(MutationRequest) (MutationResponse, error)
    // ...etc...
}

type AppProvider interface {
    Manifest() AppManifest
    NewApp(AppConfig) (App, error)
}
```

Runners then consume an AppProvider, and handle initializing the app by providing it with config data, including kube config, which are based on the runtime manner (for example, an API server would use the loopback config, or an Operator would use the config for the API server).

This requires the existence of the App Manifest as described in https://github.com/grafana/grafana-app-sdk/issues/353

PR #402, opened 2024-09-12
App Logic and Runtime Separation

IfSentient:
# What This PR Does / Why We Need It

This PR resolves https://github.com/grafana/grafana-app-sdk/issues/385 by introducing a new way to write and run apps focused around app logic implementing the interface `app.App`, and runtime logic (such as a standalone operator) existing as a layer over `app.App` which translates runtime-specific behaviors into the `app.App` method calls.

## App Logic

App logic is now handled as implementation of the new `app.App` interface, which describes all possible app behaviors. A default implementation, akin to `simple.Operator`, exists as `simple.App` (created with `simple.NewApp`). A user can also write a custom implementation of `app.App` if they so wish, and it should be compatible with any runner that consumes an `app.App`/`app.Provider`.

`app.Provider` is an interface for usage by runners, which provides an `app.Manifest` to the runner to get app capabilities, and then allows for creation of an `app.App` with configuration sourced from the runner (such as KubeConfig, which may have varying methods by which it is loaded depending on the runner).

## Runtime Logic

This PR introduces only one runner, `simple.StandaloneOperator`, which runs an app as a standalone operator. Combined with `simple.App`, this gives users the same (and slightly more) functionality that `simple.Operator` currently does. Planned in a future PR is `plugin.App` (or a similar name), which will run the app as a plugin and translate gRPC admission and CallResource calls into the app method calls.

## Codegen & Documentation

Codegen for `grafana-app-sdk component add` will be introduced in a future PR to keep this one lighter, alongside documentation and examples for writing an app in this new manner (and existing documentation will be updated to working with apps this way). This PR is focused on introducing the new functionality, and as the "old" (current) way of writing apps is still perfectly valid, documentation will be added and updated in a docs-focused PR.

## Other Changes

This PR also contains a drive-by fix for app manifest go codegen, as it currently improperly duplicates the `Validation` field for mutation when enabled.


simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+// AppInformerConfig contains configuration for the App's internal operator.InformerController
+type AppInformerConfig struct {
+	ErrorHandler       func(context.Context, error)
+	RetryPolicy        operator.RetryPolicy
+	RetryDequeuePolicy operator.RetryDequeuePolicy
+	FinalizerSupplier  operator.FinalizerSupplier
+}
+
+// AppManagedKind is a Kind and associated functionality used by an App.
+type AppManagedKind struct {
```
joeblubaugh, line 80: Is there any distinction between a kind that the app manipulates vs one that it has a read-only relationship with? Is that distinction useful?

app/app.go
```diff
@@ -0,0 +1,108 @@
...
+	// Mutate runs mutation on the incoming request, responding with a MutatingResponse on success, or an error on failure
+	Mutate(ctx context.Context, request *resource.AdmissionRequest) (*resource.MutatingResponse, error)
+	// Convert converts the object based on the ConversionRequest, returning a RawObject which MUST contain
+	// the converted bytes and encoding (Raw and Encoding respectively), and MAY contain the Object representation of those bytes.
+	// It returns an error if the conversion fails, or if the functionality is not supported by the app.
+	Convert(ctx context.Context, req ConversionRequest) (*RawObject, error)
+	// CallSubresource handles the subresource call, and writes the response to the http.ResponseWriter.
+	// If a non-http-error response is encountered, an error should be returned.
+	// It returns an error if the functionality is not supported by the app.
+	CallSubresource(ctx context.Context, writer http.ResponseWriter, req *SubresourceRequest) error
```
radiohead, line 99: QUESTION – Should we abstract the response handling here? Maybe something like:
```suggestion
	CallSubresource(ctx context.Context, req *SubresourceRequest) (RawObject, error)
```
where `RawObject` could be something like:
```golang
type RawObject interface {
	runtime.Object
	schema.ObjectKind
	metav1.Object
}
```

WDYT?

simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+	// Kind is the resource.Kind being managed. This is equivalent to a kubernetes GroupVersionKind
+	Kind resource.Kind
+	// Reconciler is an optional reconciler to run for this Kind. Only one version of a Kind should have a Reconciler,
+	// otherwise, duplicate events will be received.
+	Reconciler operator.Reconciler
+	// Watcher is an optional Watcher to run for this Kind. Only one version of a Kind should have a Watcher,
+	// otherwise, duplicate events will be received.
+	Watcher operator.ResourceWatcher
```
radiohead, line 88: This should be either–or, right? Should we consider dropping `Watcher` from here, since we've discussed moving to `Reconciler` long-term?

simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+	// Validator is an optional ValidatingAdmissionController for the Kind. It will be run only for validation
+	// of this specific version.
+	Validator resource.ValidatingAdmissionController
+	// Mutator is an optional MutatingAdmissionController for the Kind. It will be run only for mutation
+	// of this specific version.
+	Mutator resource.MutatingAdmissionController
+	// SubresourceRoutes are an optional map of subresource paths to a route handler.
+	// If supported by the runner, calls to these subresources on this particular version will call this handler.
+	SubresourceRoutes map[string]func(ctx context.Context, writer http.ResponseWriter, req *app.SubresourceRequest) error
```
radiohead, line 97: I think we want to route based on at least verb + path / subresource name – should we create a custom struct for this?

simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+	// LabelFilters are any label filters to apply to the ListWatch request
+	LabelFilters []string
+	// FieldSelectors are any field selector filters to apply to the ListWatch request
+	FieldSelectors []string
+	// UsePlain can be set to true to avoid wrapping the Reconciler or Watcher in its Opinionated variant.
+	UsePlain bool
+}
+
+// NewApp creates a new instance of App, managing the kinds provided in AppConfig.ManagedKinds.
+// AppConfig MUST contain a valid KubeConfig to be valid. Kinds can be managed by the app either with
```
radiohead, line 115: Can we validate it and return an error if validation fails?

simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// AddRunnable adds an arbitrary resource.Runnable runner to the App, which will be encapsulated as part of Runner().Run().
+// If the provided runner also implements metrics.Provider, PrometheusCollectors() will be called when called on Runner().
+func (a *App) AddRunnable(runner app.Runnable) {
+	a.operator.AddController(runner)
+}
+
+// ManageKind introduces a new kind to manage.
+func (a *App) ManageKind(kind AppManagedKind) error {
```
radiohead, line 191: What is the use-case for this method? Why would someone use this instead of passing all kinds to the `NewApp` instead?

simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+			if err != nil {
+				return fmt.Errorf("could not add watcher to controller: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
+// RegisterKindConverter adds a converter for a GroupKind, which will then be processed on Convert calls
+func (a *App) RegisterKindConverter(groupKind schema.GroupKind, converter k8s.Converter) {
```
radiohead, line 250: Same as above – why would this be used over passing all converters upon initialization?

simple/app.go
```diff
@@ -0,0 +1,334 @@
...
+	converters         map[string]Converter
+}
+
+// AppConfig is the configuration used by App
+type AppConfig struct {
+	Name           string
+	KubeConfig     rest.Config
+	InformerConfig AppInformerConfig
+	ManagedKinds   []AppManagedKind
+	Converters     map[schema.GroupKind]Converter
```
radiohead, line 68: Do you think this could be part of the `AppManagedKind` instead?

app/app.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Provider represents a type which can provide an app manifest, and create a new App when given a configuration.
+// It should be used by runners to determine an app's capabilities and create an instance of the app to run.
+type Provider interface {
+	// Manifest returns a Manifest, which may contain ManifestData or may point to a location where ManifestData can be fetched from.
+	// The runner should use the ManifestData to determine app capabilities.
+	Manifest() Manifest
+	// NewApp creates a new App instance using the provided config, or returns an error if an App cannot be instantiated.
+	NewApp(Config) (App, error)
+}
```
radiohead, line 65: SUGGESTION – I think provider should provide both config and manifest, separately. That way we can completely separate the bootstrap phase from the actual running of the app and use different bootstrap configs used by the runtime for fetching manifests & configs from configs for the apps.

```suggestion
// Config ...
type Config struct {
	// KubeConfig is a kubernetes rest.Config used to communicate with the API server where the App's Kinds are stored.
	KubeConfig rest.Config
	// ExtraConfig is additional configuration information, which may be app or runner specific
	ExtraConfig map[string]any
}

// Provider represents a type which can provide an app manifest, and create a new App when given a configuration.
// It should be used by runners to determine an app's capabilities and create an instance of the app to run.
type Provider interface {
	// Manifest returns a Manifest, which may contain ManifestData or may point to a location where ManifestData can be fetched from.
	// The runner should use the ManifestData to determine app capabilities.
	Manifest() Manifest

	// Config ...
	Config() Config

	// NewApp creates a new App instance using the provided config, or returns an error if an App cannot be instantiated.
	NewApp(Manifest, Config) (App, error)
}
```

simple/app_operator.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+type capabilities struct {
+	conversion bool
+	mutation   bool
+	validation bool
+}
+
+// Run runs the StandaloneOperator using the provided config, until stopCh is closed or receives a message, or an unrecoverable error occurs
+func (s *StandaloneOperator) Run(config OperatorAppConfig, stopCh <-chan struct{}) error {
```
radiohead, line 70: SUGGESTION – how about inverting the order of execution and requiring the confing in `NewStandaloneOperator` but passing the provider here? That would allow one to run multiple apps with one operator.

Also, we could switch to context instead of using channels.

```suggestion
func (s *StandaloneOperator) Run(ctx context.Context, provider app.Provider) error {
```

radiohead (commented):
Looks great! A few questions / suggestions.

app/app.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Runnable represents a type which can be run until it errors or the provided channel is stopped (or receives a message)
+type Runnable interface {
+	// Run runs the process and blocks until one of the following conditions are met:
+	// * An unrecoverable error occurs, in which case it returns the error
+	// * The provided channel is closed, in which case processing should stop and the method should return
+	// * The provided channel is sent a message, in which case processing should stop and the method should return
+	// * The process completes and does not need to run again
+	Run(<-chan struct{}) error
```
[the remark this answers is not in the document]
IfSentient: I'm fine with that, this was mainly modeled after the way kubernetes handles things so that the channel could be passed around internally, but we can always translate into a channel for the k8s stuff.

simple/app.go
IfSentient: Yeah, this is an either-or (if you use both, you'll get duplicated events). I'm not sure about dropping Watcher yet as it seems like most new apps prefer the Watcher approach, I think the first step might be to have the generated code prefer reconcilers.

simple/app.go
IfSentient: The payload of the request includes the verb, so I was on the fence about whether the handler should be verb-specific or agnostic. I can try introducing verb-specific handlers here and see how it feels ergonomically.

Currently, there are no runners that use custom subresource routes, so it can also be migrated in the future before we introduce them if necessary, but I'd like to get it right in this PR if possible.

simple/app.go
IfSentient: Convenience, mainly. There have been some comments on the large config object getting unwieldy, though the contents of the config vs this method are the same thing, so it's mostly about adding the option.

simple/app.go
IfSentient: I tried messing around with it, the problem being that `AppManagedKind` is a GroupVersionKind, and a converter is for a GroupKind. We'd have to restructure the way that conversion is handled, and I didn't have a good solution for it that didn't feel overly complicated. I'd still like to revisit this in the future.

simple/app.go
IfSentient: I've introduced `simple.AppUnmanagedKind` and `simple.App.WatchKind(simple.AppUnmanagedKind)` for establishing watchers/reconcilers on kinds which the app does not manage (this can also still be done via the `AddRunnable` method, but it's more cumbersome).

simple/app_operator.go
IfSentient: With how it currently works under the hood without rewriting substantial parts of code, you can't run multiple apps under the same StandalongOperator, because `Run` sets up and runs an operator that includes a `k8s.WebhookServer`, which is bound to a specific port from the config. If you tried to call `Run` again with a different (or event the same) provider, it would try to re-bind to that port (based on the config) and fail. We could have the config _and_ the provider passed to `Run`, or perhaps a "minimal" config, but at that point, it seems like it would be just as easy for a user to have two `StandalongOperator` instances.

I have converted `Run` to use a `context.Context`, though, to match `app.Runnable`.

app/app.go
radiohead: If this is the case then we need to rethink the wording and the approach - the described use case does not fit the subresource bill.

simple/app.go
radiohead: Let's make sure then that we validate that only either one is passed.

simple/app.go
radiohead: I think matching should be done in the sdk as much as possible, so let's move to verb + path.

simple/app.go
radiohead: Let's limit the options then. I see footguns potential here - due to synchronization / concurrency.

app/app.go
radiohead: I think that ideally the runner would be responsible for fetching & loading the app config from various sources (env, args, remote API) and passing it to the app.

It should definitely consist of at least two parts – general config that is known to the runner (e.g. kubeconfig, ports for listeners, API endpoints, etc. etc.) and app-specific config, opaque to the runner.

I suggest that the former should be codified here in the Provider, the latter I'm less certain about but IMHO we _could_ start by having some provider-opaque config (which could e.g. be `resource.Object` or `map[string]any`), which is part of the composite config. The app then would be responsible for properly parsing / casting the config data, maybe that parsing / casting code could be generated via codegen in the future.

simple/app_operator.go
radiohead: How about future-proofing the API? I think we should consider rewriting some code to make it possible to `Run` multiple apps under one operator, which would include having a shared webhook server listener, etc. etc.

app/app.go
IfSentient: I've introduced `app.SpecificConfig` (which right now is just an alias to `any`), which is now included in `app.Config` in place of `ExtraConfig`. `app.Provider` now also requires `SpecificConfig()` which returns the app-specific `SpecificConfig`.

simple/app_operator.go
IfSentient: OK, I didn't want to touch the underlying `k8s` code due to that breaking extant projects, and what we needed anyway was the ability to have `Run` use the same webserver, so I introduced `app.SingletonRunner`, which has a wrapped `Runnable` that multiple calls to `Run` will only run the underlying `Run` if it isn't already running (started this as specific logic in `simple.StandaloneOperator`, but it felt like it might be more useful as generic logic for any author to use). I think we'll eventually need to re-write/move/update the `k8s.WebhookServer` logic, but for now it seemed most expedient to re-use it with a little bit of a hack underneath. This ensures that he API is good, even if we need to update the underlying code later.

app/app.go
IfSentient: I've updated the method name and the response is now a more concrete object, though still not necessarily a kubernetes one.

simple/app.go
IfSentient: We can validate that it is present, but I'm not sure what to check to see if it's valid, aside from attempting to use it to talk to a known resource. I suppose we could do thatusing a resource from the manifest. Thoughts?

app/app.go
```diff
@@ -0,0 +1,122 @@
...
+// App represents an app platform application logical structure.
+// An App is typically run with a wrapper, such as simple.NewStandaloneOperator,
+// which will present a runtime layer (such as kubernetes webhooks in the case of an operator),
+// and translate those into calls to the App. The wrapper is typically also responsible for lifecycle management
+// and running the Runnable provided by Runner().
+// Pre-built implementations of App exist in the simple package, but any type which implements App
+// should be capable of being run by an app wrapper.
+type App interface {
+	// Validate validates the incoming request, and returns an error if validation fails
+	Validate(ctx context.Context, request *resource.AdmissionRequest) error
```
ferruvich, line 102: very nit: since we have `AdmissionRequest` and `MutatingResponse` re-declared above, should we use them?

IfSentient: Yes, we should be using them here, I completely forgot to do so 🤦 . Good catch.

IfSentient: OK, implementing was a bit more work than I anticipated, but `Validate` and `Mutate` now use the `app.AdmissionRequest` and `app.MutatingResponse`. Had to introduce some new interfaces and types in `simple` (`simple.KindValidator` and its simple implementation `simple.Validator`, and `simple.KindMutator` and its implementation `simple.Mutator`), and then a translation in `operator.Runner` because `k8s.WebhookServer` wants the `resource` variants. Probably future work to make this whole chain more efficient, but for now, the API for `app.App` is correct, which is what matters more.

operator/runner.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// NewRunner creates a new, properly-initialized instance of a Runner
+func NewRunner(cfg RunnerConfig) (*Runner, error) {
+	// Validate the KubeConfig by constructing a rest.RESTClient with it
+	// TODO: this requires a GroupVersion, which gets set up based on the kind
+	// _, err := rest.RESTClientFor(&cfg.KubeConfig)
+	// if err != nil {
+	// 	return nil, fmt.Errorf("invalid KubeConfig: %w", err)
+	// }
```
radiohead, line 44: I think just validating that it has things like user / token / api address / certs / etc. is enough.

I was under the impression that there was a validation function like that somewhere in `client-go` but that doesn't seem to be the case.

I suggest we leave this as is for now and write a function like that later.

app/runner.go
```diff
@@ -0,0 +1,198 @@
...
+	// ExitWait is how long to wait for Runners to exit after ErrorHandler returns true or the context is canceled
+	// before stopping execution and returning a timeout error instead of exiting gracefully.
+	// If ExitWait is nil, Run execution will always block until all Runners have exited.
+	ExitWait *time.Duration
+}
+
+// Run runs all Runners in separate goroutines, and calls ErrorHandler if any of them exits early with an error.
+// If ErrorHandler returns true (or if there is no ErrorHandler), the other Runners are canceled and the error is returned.
+func (m *MultiRunner) Run(ctx context.Context) error {
+	propagatedContext, cancel := context.WithCancel(ctx)
```
radiohead, line 45: NIT – I'm not 100% sure but I think most of this can be replaced with https://pkg.go.dev/golang.org/x/sync/errgroup.

radiohead (commented):
Looks great! Just a few last bits.

app/app.go
radiohead: Nice, thank you!

simple/app_operator.go
radiohead: Nice, thank you!

simple/app.go
radiohead: I left another comment – I thought there was a client-side only validation function in the client-go but turns out I was wrong. Let's maybe write something like that but it can be done in a follow-up PR(s), later on.

app/app.go
radiohead: Looks much cleaner, thank you!

simple/app.go
```diff
@@ -88,19 +132,37 @@ type AppManagedKind struct {
...
 	// ReconcileOptions are the options to use for running the Reconciler or Watcher for the Kind, if one exists.
-	ReconcileOptions ReconcileOptions
+	ReconcileOptions BasicReconcileOptions
 }
 
-// ReconcileOptions are settings for the ListWatch and informer setup for a reconciliation loop
-type ReconcileOptions struct {
+type AppCustomRouteHandler struct {
+	Path    string
+	Methods []string
```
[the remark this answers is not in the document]
IfSentient: That works, I was just using similar syntax to a lot of popular routers that let you attach multiple methods to a handler. We could always change this in the future if we need to as well, as this is the part that can evolve since there are no current runners that support custom routes.

simple/app.go
```diff
@@ -88,19 +132,37 @@ type AppManagedKind struct {
...
+	Validator KindValidator
 	// Mutator is an optional MutatingAdmissionController for the Kind. It will be run only for mutation
 	// of this specific version.
-	Mutator resource.MutatingAdmissionController
-	// SubresourceRoutes are an optional map of subresource paths to a route handler.
+	Mutator KindMutator
+	// CustomRoutes are an optional map of subresource paths to a route handler.
 	// If supported by the runner, calls to these subresources on this particular version will call this handler.
-	SubresourceRoutes map[string]func(ctx context.Context, writer http.ResponseWriter, req *app.SubresourceRequest) error
+	CustomRoutes []AppCustomRouteHandler
```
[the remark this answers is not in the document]
IfSentient: That makes sense to me, let me implement this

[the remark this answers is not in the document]
IfSentient: So it looks like using a map doesn't forbid duplicate entries, so I still have the duplicates check. Given this, does having this be a slice of 
```type AppCustomRoute struct {
    Method AppCustomRouteMethod
    Path string
    Handler AppCustomRouteHandler
}```
Make more sense? I'll commit the version with the map for now.

File: app/runner.go (hunk 14 above)
Comment (radiohead) [reply app/runner.go:45]: > It looks like errgroup.Group will always wait for all goroutines to complete, so if we want to exit early on an error, we'd still need to track and close all the other ones.

It will indeed wait for all goroutines to complete, but it will cancel the internal context as soon as any goroutine returns an error, giving us the ability to either exit as soon as any goroutine returns an error OR the parent context is canceled. In both cases the benefit is that all underlying goroutines will be canceled early and ensured that they will run to completion.

The only difference would be that we will lose the ability to exit early without waiting for all goroutines to return, within a given timeout, but I'd argue this logic should only exist in the entrypoint of the app (i.e. the `main` function).

> We can't remove the errors channel in favor of it, either, because we need to be able to possibly respond differently based on the error.

Since `g.Wait` bubbles up the error from the underlying goroutine it already handles that for us, no?

File: simple/app.go (hunk 16 above)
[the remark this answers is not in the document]
Comment (radiohead) [reply simple/app.go]: > So it looks like using a map doesn't forbid duplicate entries as a compiler error, but instead just overwrites them.

Good point, but at least in this case it will be on the developer. Assuming that they declare the map in-line, the compiler will fail, since it can statically determine duplicate entries. If the developer builds the route map dynamically they will still have the ability to shoot themselves in the foot but to prevent that we could in the future change the custom type to something that uses the map internally but doesn't expose the map directly, instead having methods like `Add` / `Delete`.

WDYT?

File: simple/app.go (hunk 15 above)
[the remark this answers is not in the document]
Comment (radiohead) [reply simple/app.go]: > That works, I was just using similar syntax to a lot of popular routers that let you attach multiple methods to a handler.

Yeah, I get where you were coming from, I just personally think that it usually is not what devs want. Plus, it's always easier to relax the requirements down the line than to increase them.

File: simple/app.go [hunk 17]
```diff
@@ -173,6 +168,24 @@ type BasicReconcileOptions struct {
 	UsePlain bool
 }
 
+type AppCustomRouteMethod string
+
+const (
+	AppCustomRouteMethodGet    AppCustomRouteMethod = "GET"
+	AppCustomRouteMethodPut    AppCustomRouteMethod = "PUT"
+	AppCustomRouteMethodPost   AppCustomRouteMethod = "POST"
+	AppCustomRouteMethodDelete AppCustomRouteMethod = "DELETE"
```
radiohead, line 177: I'm really sorry for not being clear in my comment, but we probably need to expand this list to cover all methods we expect to expose to developers.

I'm not sure if we need the likes of `HEAD` or `CONNECT` but we at least should include `PATCH`.

Maybe just copying what `net/http` has is enough:

```suggestion
	AppCustomRouteMethodConnect AppCustomRouteMethod = http.MethodConnect
	AppCustomRouteMethodDelete  AppCustomRouteMethod = http.MethodDelete
	AppCustomRouteMethodGet     AppCustomRouteMethod = http.MethodGet
	AppCustomRouteMethodHead    AppCustomRouteMethod = http.MethodHead
	AppCustomRouteMethodOptions AppCustomRouteMethod = http.MethodOptions
	AppCustomRouteMethodPatch   AppCustomRouteMethod = http.MethodPatch
	AppCustomRouteMethodPost    AppCustomRouteMethod = http.MethodPost
	AppCustomRouteMethodPut     AppCustomRouteMethod = http.MethodPut
	AppCustomRouteMethodTrace   AppCustomRouteMethod = http.MethodTrace
```

simple/app.go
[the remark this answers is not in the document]
IfSentient: > Assuming that they declare the map in-line, the compiler will fail, since it can statically determine duplicate entries
That's what I mean--it doesn't run into a compile-time error, the map keys just overwrite each other even when in-line ([playground example](https://go.dev/play/p/Yar8q1QYMHM)). Not sure if this is desirable UX, as a user may think (like I did initially) that if they in-line declare the map, they're protected from duplicates by the compiler.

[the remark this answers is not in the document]
radiohead: Oh, I see now, must be that it only works for scalar types. Anyhow, I think it should be OK for now – like I said, we can introduce a new custom wrapper over this map in later PRs.

radiohead (approved):

PR #405, opened 2024-09-13
[Integration tests]: manually get the current commit for the action when bootstrapping the project.

IfSentient:
This PR does two things:
* Moves testing to a reusable workflow which is now shared between the `main`, `pr`, and `release` workflows.
* Updates the integration tests to use the version of the grafana-app-sdk that matches the code the action is being run against (using a `replace` directive in the `go.mod`, as using `GITHUB_SHA` doesn't cooperate with `go get` in PR builds).

Currently, the branch protection settings for `main` require that an action named `test` complete successfully to merge (which no longer exists the same way, so this blocks indefinitely), once this PR is approved I will update the branch protection settings to require `test / unit-test`, `test / integration-test`, and `test / build` instead, then merge this PR so other PR's will need to merge in from main to get correct testing.


radiohead (commented):
@IfSentient Looks like using `git` to get commit rev won't work, but we could get that info from the env vars – https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables.

IIUC `GITHUB_SHA` should fit the bill.

IfSentient:
> @IfSentient Looks like using `git` to get commit rev won't work, but we could get that info from the env vars – https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables.
> 
> IIUC `GITHUB_SHA` should fit the bill.

I tried that before, and I get the error
```
go: github.com/grafana/grafana-app-sdk@9bb57268ee194cd7cfa69a892ec20d1293134cef: invalid version: unknown revision 9bb57268ee194cd7cfa69a892ec20d1293134cef
```
It looks like [9bb57268ee194cd7cfa69a892ec20d1293134cef](https://github.com/grafana/grafana-app-sdk/commit/9bb57268ee194cd7cfa69a892ec20d1293134cef) exists, but is some kind of fake commit that is the totality of the PR instead of being a commit in the branch or in main, which I suspect is why `go get` is having problems.

radiohead (approved):

PR #419, opened 2024-09-30
[operator] Add the ability to customize and disable the cache resync interval of the KubernetesBasedInformer

IfSentient:
This PR updates the `operator.NewKubernetesBasedInformerWithFilters` method to use a config object that includes both ListWatchOptions and a CacheResyncInterval duration, which is passed to the underlying `cache.NewSharedIndexInformer` call, allowing for customization of the cache resync interval instead of the default 30s that was hard-coded (and set to 30s due to an ancient copy/paste).

This value can also be set via `simple.NewOperator` with `simple.OperatorConfig.InformerCacheResyncInterval`.


radiohead (approved):

ferruvich (approved):
LGTM!

PR #423, opened 2024-10-02
[codegen][docs] Update New Project Codegen and Tutorial Doc

IfSentient:
This PR updates the new project codegen from `grafana-app-sdk project component add operator` to use `simple.App` and `operator.Runner`, as opposed to `simple.Operator`. It now splits the app out into two files: `/cmd/operator/main.go`, which contains the loading of config, and running the operator runner, and `/pkg/app/app.go`, which contains the app logic, including the Provider and NewApp functionality. This allows the app to be re-used for other runners if desired, though other `/cmd/*` packages.

This PR also updates the tutorial based on these changes, so that it follows along with the new way of writing apps and reflects the codegen as of this PR.

Finally, there is also an update to the project local generation, used with `grafana-app-sdk project local generate`, which now no longer uses values in the config for admission/conversion capabilities and instead uses the capabilities declared in the app manifest.


codegen/templates/operator/main.tmpl
```diff
@@ -37,59 +35,51 @@ func main() {
...
     if err != nil {
-        logging.DefaultLogger.With("error", err).Error("Error adding {{.Kind}} watcher to controller")
+        logging.DefaultLogger.With("error", err).Error("Unable to create operator runner")
         panic(err)
     }
-    {{ end }}{{ end }}
 
-    stopCh := make(chan struct{})
+    // Context and cancel for the operator's Run method
+    ctx, cancel := context.WithCancel(context.Background())
```
radiohead, line 69: Very small nit, but since we've moved to contexts from raw channels, how about using:
```go
ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill)
```
and getting rid of the low-level signal handling and extra goroutine?

radiohead (approved):

radiohead (approved):

Issue #271, opened 2024-04-30
[local] Enable k3d API Server Access via Grafana API Server in Local Dev Environment


IfSentient:
In order to enable direct access to the API server, we need a way for the front-end grafana plugin to communicate directly with the API server using grafana auth. The likely most feasible way to do this (at least initially) is to have the k3d API extensions server (which serves CRD's) aggregated under the grafana API server, so that a front-end plugin can talk to the grafana API server to access their CRDs.

We need a solution that works with the manifests generated by `grafana-app-sdk project local generate` that will enable a front-end plugin to take the same actions on their kinds as the back-end plugin currently can, without needing that back-end proxy.

toddtreece:
there are [docs](https://github.com/grafana/grafana/tree/main/pkg/services/apiserver/aggregator#testing-aggregation-locally) on how to configure aggregation, and [examples](https://github.com/grafana/grafana/tree/main/pkg/services/apiserver/aggregator/examples/manual-test) of what to apply. let me know if that will cover what you need

PR #446, opened 2024-10-29
[Local Environment Setup] Automatically Aggregate App CRDs Under Grafana API Server

IfSentient:
As part of the `grafana-app-sdk project local generate`, create a new script to aggregate app CRDs under a grafana API server. Additionally, add a tilt step in the new-project generated tilt file to do this aggregation automatically.

Resolves https://github.com/grafana/grafana-app-sdk/issues/271


radiohead (approved):
Nice feature. Looks very messy, but let's merge it as-is. Ideally though, it might be a good idea to think about how we can better handle project generation and specifically simplify aggregation in the future.

radiohead:
@IfSentient I assume this closes https://github.com/grafana/grafana-app-sdk/issues/271?

PR #447, opened 2024-10-30
[codegen] Remove Thema Support

IfSentient:
Remove thema support for codegen, and deprecate model type generation for a future removal once kind definition is moved into manifest definition.


radiohead (approved):
:shipit:

PR #452, opened 2024-11-04
Add LTS branch build workflow and have tests run on PR's to LTS branch

IfSentient:
PR tests aren't currently run for the LTS branch, fix that.

### DIFF
--- a/.github/workflows/lts.yml
+++ b/.github/workflows/lts.yml
@@ -1,0 +1,10 @@
+name: Build Main
+
+on:
+  push:
+    branches:
+      - "lts/v0.24"
+
+jobs:
+  test:
+    uses: ./.github/workflows/test.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/lts.yml
```diff
@@ -0,0 +1,10 @@
+name: Build Main
+
+on:
+  push:
+    branches:
+      - "lts/v0.24"
```
radiohead, line 6: I _think_ this can be `lts/*` instead?

.github/workflows/pr.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
     - main
+    - "lts/v0.24"
```
radiohead, line 7: Same here.

radiohead (approved):
LGTM, but we might want to run these on all LTS branches.

radiohead (approved):

PR #462, opened 2024-11-06
Use full branch name for LTS branch, without a wildcard.

IfSentient:
It looks like the workflows aren't running on a PR to the LTS branch, so try using the full name instead of a wildcard.


radiohead (approved):

IfSentient:
Gonna merge this to see if it fixes the problem, but [a fork with the same GH workflows works correctly](https://github.com/IfSentient/grafana-app-sdk/pull/3), so this needs more investigation on my end

PR #478, opened 2024-11-12
[webhooks] Support Version Distinction in Admission Control

IfSentient:
Have the k8s.WebhookServer use GroupVersionKind when mapping validating and mutating controllers, rather than GroupKind. The methods for adding admission control already use `resource.Kind`, which is a GVK-scoped object, so just support this when storing in the relation map and lookups for handling requests.


radiohead (approved):
Nice!

PR #481, opened 2024-11-14
[codegen] Add Webhook Conversion Strategy to Generated CRDs

IfSentient:
Add extra optional webhook conversion params to apiResource in CUE kind, use those parameters to add webhook conversion strategy to generated CRD's when apiResource.conversion = true.

The extra parameters are a temporary measure as we move toward the codegen having a manifest to look at which would include the necessary information.


codegen/cuekind/def.cue
```diff
@@ -142,6 +142,13 @@ Kind: S={
 		}
 		// conversion determines whether there is code-based conversion for this kind. Used for generating the manifest.
 		conversion: bool | *false
+		// conversionWebhookProps is a temporary way of specifying the service webhook information
+		// which will be migrated away from once manifests are used in the codegen pipeline
+		conversionWebhookProps: {
+			serviceName: string | *""
+			serviceNamespace: string | *""
```
radiohead, line 149: We need to support `url` (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#webhookclientconfig-v1-apiextensions-k8s-io) here, either as an option, or as the main & only way – we don't use service configs at all.

IfSentient: I'll change it to the main & only way, as this is something we'll likely want to address better in the manifest and this is a temporary patch.

radiohead: SGTM!

There's an issue with the code at the moment, since it doesn't comply with CRD schema – the config should be either a Service or an URL.

I suggest that instead of hand-rolling custom types, we instead use the canonical ones from upstream, e.g. something like this:

```diff
diff --git a/codegen/jennies/crd.go b/codegen/jennies/crd.go
index 17c5f91..32ad450 100644
--- a/codegen/jennies/crd.go
+++ b/codegen/jennies/crd.go
@@ -3,11 +3,13 @@ package jennies
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	"cuelang.org/go/cue"
 	"github.com/grafana/codejen"
 	goyaml "gopkg.in/yaml.v3"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
 	"github.com/grafana/grafana-app-sdk/codegen"
 	"github.com/grafana/grafana-app-sdk/k8s"
@@ -53,15 +55,18 @@ func (c *crdGenerator) Generate(kind codegen.Kind) (*codejen.File, error) {
 	}
 
 	if kind.Properties().APIResource.Conversion {
-		resource.Spec.Conversion = &k8s.CustomResourceDefinitionSpecConversion{
-			Strategy: "webhook",
-			Webhook: &k8s.CustomResourceDefinitionSpecConversionWebhook{
+		u := kind.Properties().APIResource.ConversionWebhookProps.URL
+
+		if _, err := url.Parse(u); err != nil {
+			return nil, err
+		}
+
+		resource.Spec.Conversion = &apiextensionsv1.CustomResourceConversion{
+			Strategy: apiextensionsv1.WebhookConverter,
+			Webhook: &apiextensionsv1.WebhookConversion{
 				ConversionReviewVersions: []string{"v1"},
-				ClientConfig: k8s.CustomResourceDefinitionClientConfig{
-					Service: k8s.CustomResourceDefinitionClientConfigService{
-						Path: kind.Properties().APIResource.ConversionWebhookProps.Path,
-						URL:  kind.Properties().APIResource.ConversionWebhookProps.URL,
-					},
+				ClientConfig: &apiextensionsv1.WebhookClientConfig{
+					URL: &u,
 				},
 			},
 		}
diff --git a/k8s/manager.go b/k8s/manager.go
index 4cabd17..cf6a4c3 100644
--- a/k8s/manager.go
+++ b/k8s/manager.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	kschema "k8s.io/apimachinery/pkg/runtime/schema"
@@ -324,32 +325,11 @@ func (crd *CustomResourceDefinition) DeepCopyObject() runtime.Object {
 
 // CustomResourceDefinitionSpec is the body or spec of a kubernetes Custom Resource Definition
 type CustomResourceDefinitionSpec struct {
-	Group      string                                  `json:"group" yaml:"group"`
-	Versions   []CustomResourceDefinitionSpecVersion   `json:"versions" yaml:"versions"`
-	Names      CustomResourceDefinitionSpecNames       `json:"names" yaml:"names"`
-	Conversion *CustomResourceDefinitionSpecConversion `json:"conversion,omitempty" yaml:"conversion,omitempty"`
-	Scope      string                                  `json:"scope" yaml:"scope"`
-}
-
-type CustomResourceDefinitionSpecConversion struct {
-	Strategy string                                         `json:"strategy" yaml:"strategy"`
-	Webhook  *CustomResourceDefinitionSpecConversionWebhook `json:"webhook,omitempty" yaml:"webhook,omitempty"`
-}
-
-type CustomResourceDefinitionSpecConversionWebhook struct {
-	ConversionReviewVersions []string                             `json:"conversionReviewVersions" yaml:"conversionReviewVersions"`
-	ClientConfig             CustomResourceDefinitionClientConfig `json:"clientConfig" yaml:"clientConfig"`
-}
-
-type CustomResourceDefinitionClientConfig struct {
-	Service CustomResourceDefinitionClientConfigService `json:"service" yaml:"service"`
-}
-
-type CustomResourceDefinitionClientConfigService struct {
-	URL       string `json:"url,omitempty" yaml:"url,omitempty"`
-	Name      string `json:"name,omitempty" yaml:"name,omitempty"`
-	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
-	Path      string `json:"path" yaml:"path"`
+	Group      string                                    `json:"group" yaml:"group"`
+	Versions   []CustomResourceDefinitionSpecVersion     `json:"versions" yaml:"versions"`
+	Names      CustomResourceDefinitionSpecNames         `json:"names" yaml:"names"`
+	Conversion *apiextensionsv1.CustomResourceConversion `json:"conversion,omitempty" yaml:"conversion,omitempty"`
+	Scope      string                                    `json:"scope" yaml:"scope"`
 }
 
 // CustomResourceDefinitionSpecVersion is the representation of a specific version of a CRD, as part of the overall spec
```

(I've also thrown in URL validation but feel free to drop it).

IfSentient: Sorry, this comment totally slipped through the cracks for me, just noticed it today. The problem with using the kubernetes types is that [they don't have yaml tags](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go), which presented problems in the past for YAML marshal (which is why the SDK has ones which are hand-rolled). I can fix the structure, though.

IfSentient: @radiohead fixed the URL portion, I think we can move over to using the k8s types if we change our YAML library, I'll try to get that working in a different PR.

radiohead: > The problem with using the kubernetes types is that [they don't have yaml tags](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go).

Ah yes, good point. I just did some digging around and it turns out k8s folks use a custom library that uses JSON tags (since YAML is a superset of JSON).

I think this is also relevant for #504 and #505 – those have been merged now but I wonder if we should revert them (maybe at a later point).

@bcotton FYI!

codegen/kind.go
```diff
@@ -36,11 +36,17 @@ type KindProperties struct {
...
+	Scope                  string                      `json:"scope"`
+	Validation             KindAdmissionCapability     `json:"validation"`
+	Mutation               KindAdmissionCapability     `json:"mutation"`
+	Conversion             bool                        `json:"conversion"`
+	ConversionWebhookProps ConversionWebhookProperties `json:"conversionWebhookProps"`
+}
+
+type ConversionWebhookProperties struct {
+	URL  string `json:"url"`
+	Path string `json:"path"`
```
radiohead, line 49: NIT – I think this is redundant, since it's not used anywhere.

radiohead (approved):

Issue #473, opened 2024-11-08
Bootstrapped Front-End Build Fails on Linux for node > 18


IfSentient:
When building the front-end generated by `grafana-app-sdk project component add frontend` on linux with a node version > 18, the following error is encountered:
```
$ yarn build
[webpack-cli] TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at exports.prepare (/home/apond/test/issue-tracker-project/plugin/.yarn/cache/rechoir-npm-0.8.0-fb660b3bc8-ad3caed8af.zip/node_modules/rechoir/index.js:21:14)
    at loadConfigByPath (/home/apond/test/issue-tracker-project/plugin/.yarn/__virtual__/webpack-cli-virtual-cc63e22c35/0/cache/webpack-cli-npm-5.1.4-7be5b53b38-3a4ad0d034.zip/node_modules/webpack-cli/lib/webpack-cli.js:1377:29)
    at /home/apond/test/issue-tracker-project/plugin/.yarn/__virtual__/webpack-cli-virtual-cc63e22c35/0/cache/webpack-cli-npm-5.1.4-7be5b53b38-3a4ad0d034.zip/node_modules/webpack-cli/lib/webpack-cli.js:1460:88
    at Array.map (<anonymous>)
    at WebpackCLI.loadConfig (/home/apond/test/issue-tracker-project/plugin/.yarn/__virtual__/webpack-cli-virtual-cc63e22c35/0/cache/webpack-cli-npm-5.1.4-7be5b53b38-3a4ad0d034.zip/node_modules/webpack-cli/lib/webpack-cli.js:1460:68)
    at WebpackCLI.createCompiler (/home/apond/test/issue-tracker-project/plugin/.yarn/__virtual__/webpack-cli-virtual-cc63e22c35/0/cache/webpack-cli-npm-5.1.4-7be5b53b38-3a4ad0d034.zip/node_modules/webpack-cli/lib/webpack-cli.js:1781:33)
    at WebpackCLI.runWebpack (/home/apond/test/issue-tracker-project/plugin/.yarn/__virtual__/webpack-cli-virtual-cc63e22c35/0/cache/webpack-cli-npm-5.1.4-7be5b53b38-3a4ad0d034.zip/node_modules/webpack-cli/lib/webpack-cli.js:1877:31)
[...]

PR #491, opened 2024-11-20
[CLI][codegen] Use `yarn create @grafana/plugin` to Generate Front-End Code

IfSentient:
Instead of using a snapshot of the output of `yarn create @grafana/plugin` with some small tweaks when adding the `frontend` component with `grafana-app-sdk project component add`, invoke the `yarn` command directly, then tweak the output as needed. This keeps us from falling behind the current plugin bootstrapping. As yarn is required to build the frontend already, we have the component add check if yarn is installed, and error if not, requesting that the user must have yarn installed to add the frontend component.

Resolves https://github.com/grafana/grafana-app-sdk/issues/473


cmd/grafana-app-sdk/project.go
```diff
@@ -609,34 +611,109 @@ func projectAddPluginAPI[G anyGenerator](generator G, repo, generatedAPIModelsPa
...
+	})
+}
+
+func isCommandInstalled(command string) bool {
+	cmd := exec.Command("which", command)
+	err := cmd.Run()
+	return err == nil
 }
 
 func writePluginJSON(fullPath, id, name, author, slug string) error {
```
radiohead, line 719: Could we also stop using templated `plugin.json` as well and instead read the one we have generated and overwrite specific fields we want to tweak?

We could even use https://github.com/grafana/grafana/blob/main/pkg/plugins/plugins.go#L137-L191, possibly.

radiohead (approved):
Looks good but I left a few suggestions, feel free to merge anyway.

cmd/grafana-app-sdk/project.go
```diff
@@ -609,34 +611,109 @@ func projectAddPluginAPI[G anyGenerator](generator G, repo, generatedAPIModelsPa
...
 	if pluginID == "" {
 		return fmt.Errorf("plugin-id is required")
 	}
 
-	err := writeStaticFrontendFiles(filepath.Join(projectRootPath, "plugin"), promptForOverwrite)
+	if !isCommandInstalled("yarn") {
+		return fmt.Errorf("yarn must be installed to add the frontend component")
+	}
+
+	args := []string{"create", "@grafana/plugin", "--pluginType=app", "--hasBackend=true", "--pluginName=tmp", "--orgName=tmp"}
```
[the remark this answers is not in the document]
IfSentient: At this point, I was just trying to mimic how the current frontend codegen acts (which always adds a backend), with the intention of cleaning this up in a future PR (because this is sort of a bug).

cmd/grafana-app-sdk/project.go
IfSentient: Yup, I'd like to update this behavior as well, but I think I'd like to have that be a future PR for the same reason as the other comment--for this PR, I'm trying to not change too much in one go, and leave the current behavior more intact.

PR #537, opened 2024-12-10
[k8s] Update KindNegotiatedSerializer to handle v1/Status objects, add logging

IfSentient:
This PR updates `k8s.KindNegotiatedSerializer` to handle kubernetes `v1/Status` objects even when not set as `into` in the `Decode` call. Debug logging has also been introduced to the decode process, and rudimentary `resource.ListObject` type handling has been added as well.

Tests have been added to check all behaviors of the Decode process.

### DIFF
--- a/k8s/negotiator.go
+++ b/k8s/negotiator.go
@@ -113,6 +113,7 @@ func (k *KindNegotiatedSerializer) SupportedMediaTypes() []runtime.SerializerInf
 	for encoding, codec := range k.Kind.Codecs {
 		serializer := &CodecDecoder{
 			SampleObject: k.Kind.ZeroValue(),
+			SampleList:   k.Kind.ZeroListValue(),
 			Codec:        codec,
 		}
 		info := runtime.SerializerInfo{
@@ -153,8 +154,14 @@ func (*KindNegotiatedSerializer) DecoderToVersion(d runtime.Decoder, _ runtime.G
 // CodecDecoder implements runtime.Serializer and works with Untyped* objects to implement runtime.Object
 type CodecDecoder struct {
 	SampleObject resource.Object
+	SampleList   resource.ListObject
 	Codec        resource.Codec
 	Decoder      func([]byte, any) error
+}
+
+type indicator struct {
+	metav1.TypeMeta `json:",inline"`
+	Items           []any `json:"items,omitempty"`
 }
 
 // Decode decodes the provided data into UntypedWatchObject or UntypedObjectWrapper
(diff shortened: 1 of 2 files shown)


k8s/negotiator.go
```diff
@@ -152,36 +154,101 @@ func (*KindNegotiatedSerializer) DecoderToVersion(d runtime.Decoder, _ runtime.G
...
+	metav1.TypeMeta `json:",inline"`
+	Items           []any `json:"items,omitempty"`
+}
+
 // Decode decodes the provided data into UntypedWatchObject or UntypedObjectWrapper
 //
 //nolint:gocritic,revive
 func (c *CodecDecoder) Decode(data []byte, defaults *schema.GroupVersionKind, into runtime.Object) (
 	runtime.Object, *schema.GroupVersionKind, error) {
 	if into != nil {
```
radiohead, line 172: Small nit – this could be changed to `switch / case` for readability.

k8s/negotiator.go
```diff
@@ -152,36 +154,101 @@ func (*KindNegotiatedSerializer) DecoderToVersion(d runtime.Decoder, _ runtime.G
...
+	if tm.Items != nil {
+		logging.DefaultLogger.Debug("decoding into a new empty list instance from kind", "gvk", tm.GroupVersionKind().String())
+		var obj resource.ListObject
+		if c.SampleList != nil {
+			obj = c.SampleList.Copy()
+		} else {
+			logging.DefaultLogger.Warn("no SampleObject set in CodecDecoder, using *resource.TypedList[*resource.UntypedObject]")
+			obj = &resource.TypedList[*resource.UntypedObject]{}
+		}
+		// TODO: use codec for each element in the list?
```
radiohead, line 237: This is my concern too – we are effectively decoding a list 2 times, which is extra CPU and memory.

I wonder if we can change `items` to be a `json.RawMessage` instead – that way we can keep current simple implementation (i.e. `err = c.Decoder(data, &obj)`) but avoid extra unmarshaling.

radiohead (commented):
Looks great, thank you! Just one note on unmarshaling items in the lists.

radiohead (approved):

k8s/negotiator.go
IfSentient: I have added a `noAlloc` type that always acts as if it is correctly unmarshaled, but contains no data. Using a `*noAlloc` pointer for the `Items` field now, so it being non-nil can be used to check if the `items` key is present in the JSON payload.

radiohead (approved):

PR #543, opened 2024-12-11
[codegen] Propagate ExtraPermissions into generated manifest data

IfSentient:
App manifest generation doesn't correctly propagate the `extraPermissions` field into the generated JSON or YAML. Fix this in the jenny, and add a test to ensure propagation is happening.


codegen/cuekind/testing/testkind.cue
```diff
@@ -5,6 +5,13 @@ import "time"
 testManifest: {
 	appName: "test-app"
 	kinds: [testKind, testKind2]
+	extraPermissions: {
+		accessKinds: [{
+			group: "foo.bar"
+			resource: "foos"
+			actions: ["get","list","watch"]
+		}]
+	}
```
radiohead, line 14: Had a chat with Kalle yesterday, concerning subresource permissions.

I think it might make sense to future-proof the design and change this to something along the lines of:
```cuelang
extraPermissions: {
	accessKinds: [
		// Use the group as the key
		"foo.grafana.app": {
			resources: {
				// Use the resource name as the key
				"foos": {
					actions: [
						"get",
						"list",
						"watch",
					],
				},
				// Use the subresource name as the key
				"foos/status": {
					actions: [
						"get",
					]
				}
			}
		}
	]
}
```

WDYT?

radiohead (approved):
Good catch!

I left a question about the future design, but it's not blocking this PR.

codegen/cuekind/testing/testkind.cue
IfSentient: Makes sense. I think it makes the simple case a bit more verbose, but it prevents other footguns/conflicts without needing manual checking.

Issue #559, opened 2024-12-19
Informer controller should have an exponential backoff in case of misconfigured finalizer


charandas:
I ran into a situation with Playlists, when switching from the OpinionatedWatcher implementation to OpinionatedReconciler, wherein I had a resource already saved with a finalizer named like below:

```yaml
apiVersion: v1
items:
- apiVersion: playlist.grafana.app/v0alpha1
  kind: Playlist
  metadata:
    finalizers:
    - playlist-playlists-finalizer
```

When I ran the OpinionatedReconciler version with the finalizer not matching the existing, the InformerController code repeatedly attempted to perform [these lines](https://github.com/grafana/grafana-app-sdk/blob/main/operator/reconciler.go#L184-L192) and propagating the NotFound error in a way that I saw the below logs, which are misleading:

```
DEBUG[12-19|12:17:25] grafana-aggregator: "/apis/playlist.grafana.app/v0alpha1/namespaces/stacks-1/playlists/adkj6k6" satisfied by NotFoundHandler logger=grafana-apiserver
```

Perhaps we can improve the logic in 2 ways:

1. Warn the user about the misconfiguration, one can only hope that they don't have mixed objects at this opinion with 2 sets of finalizer values
2. Alternatively, offer a way of fixing the issue by resorting to the newer value for existing objects - not sure how to intelligently do so without user hint

Additionally, we need to fix the infinite try loop that is sure to peg the processor:

1. Do this by performing the retry logic within limits with exponential backoff

IfSentient:
Errors returned from a reconciler should [be retried based on the RetryPolicy](https://github.com/grafana/grafana-app-sdk/blob/main/operator/informer_controller.go#L564) for the InformerController, which, by default, is an [Exponential Backoff policy](https://github.com/grafana/grafana-app-sdk/blob/main/operator/informer_controller.go#L35). The only way I see a tight loop happening here is a bad RetryPolicy, or (more likely), continuous reconcile events triggered by patches (in which case, the patch would have been successful, and the reconciler _should_ have been removed?)

[...]

PR #560, opened 2025-01-02
[OpinionatedReconciler] Add debug logging and tracing, change some logic to avoid retry loops on deletes

IfSentient:
Related to https://github.com/grafana/grafana-app-sdk/issues/559, added debug logging and tracing similar to what `OpinionatedWatcher` already has. While testing, changed the logic in the `OpinionatedReconciler` to resolve two issues:
1. Previously, when an update added the `deletionTimestamp`, the `OpinionatedReconciler` would remove the finalizer and return, and then rely on the subsequent delete event to be propagated down to the downstream reconciler for delete behavior. This introduces an edge case that is addressed in `OpinionatedWatcher` but missed here, where, if the downstream's processing of the delete fails, the object is still deleted in the API server, possibly leaving a bad state (in `OpinionatedWatcher`, the finalizer is not removed until the downstream delete is successfully processed). The logic has now been changed to mimic that of the `OpinionatedWatcher`, sending the "add `deletionTimestamp`" event to the downstream reconciler as a delete event, and only removing the finalizer once this is successful. As a side effect, _actual_ delete events are no long propagated to the downstream.
2. After the final finalizer is removed from an object with a non-nil `deletionTimestamp`, a delete event is emitted in the watch stream which _still contains_ the last finalizer to be removed (if there were multiple finalizers, only the last one is present in the list, i.e. the last state the object was in before the finalizers list became empty), rather than an empty finalizers list. To combat having this make us attempt to then remove the finalizer _again_ (because the object is actually deleted now and this will result in an error which will be retried by default), the swallowing of delete events has been moved to _before_ any of the other deletionTimestamp checks.

Example of deletion finalizers list behavior from (2):
```
% kubectl get --raw '/apis/issuetrackerproject.ext.grafana.com/v1/namespaces/default/issues?watch=1'
{"type":"ADDED","object":{"apiVersion":"issuetrackerproject.ext.grafana.com/v1","kind":"Issue","metadata":{"creationTimestamp":"2025-01-02T17:46:29Z","generation":1,...trimmed...}}}
{"type":"MODIFIED","object":{"apiVersion":"issuetrackerproject.ext.grafana.com/v1","kind":"Issue","metadata":{"creationTimestamp":"2025-01-02T17:46:29Z","finalizers":["issue-tracker-project-issues-finalizer"],"generation":1,...trimmed...}}}
{"type":"MODIFIED","object":{"apiVersion":"issuetrackerproject.ext.grafana.com/v1","kind":"Issue","metadata":{"creationTimestamp":"2025-01-02T17:46:29Z","finalizers":["issue-tracker-project-issues-finalizer","foo"],"generation":1,...trimmed...}}}
{"type":"MODIFIED","object":{"apiVersion":"issuetrackerproject.ext.grafana.com/v1","kind":"Issue","metadata":{"creationTimestamp":"2025-01-02T17:46:29Z","deletionGracePeriodSeconds":0,"deletionTimestamp":"2025-01-02T17:47:14Z","finalizers":["issue-tracker-project-issues-finalizer","foo"],"generation":2,...trimmed...}}}
{"type":"MODIFIED","object":{"apiVersion":"issuetrackerproject.ext.grafana.com/v1","kind":"Issue","metadata":{"creationTimestamp":"2025-01-02T17:46:29Z","deletionGracePeriodSeconds":0,"deletionTimestamp":"2025-01-02T17:47:14Z","finalizers":["foo"],"generation":2,...trimmed...}}}
{"type":"DELETED","object":{"apiVersion":"issuetrackerproject.ext.grafana.com/v1","kind":"Issue","metadata":{"creationTimestamp":"2025-01-02T17:46:29Z","deletionGracePeriodSeconds":0,"deletionTimestamp":"2025-01-02T17:47:14Z","finalizers":["foo"],"generation":2,...trimmed...}}}
```
Here a new `issue` is created (from the tutorial project), the reconciler adds its finalizer, then an extra `foo` finalizer is added. The object is deleted, and the reconciler automatically removes its finalizer. The `foo` finalizer is removed, but the state of the object doesn't change and the delete happens with that state.

### DIFF
--- a/operator/reconciler.go
+++ b/operator/reconciler.go
@@ -5,8 +5,11 @@ import (
 	"fmt"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	"k8s.io/utils/strings/slices"
 
+	"github.com/grafana/grafana-app-sdk/logging"
 	"github.com/grafana/grafana-app-sdk/resource"
 )
 
@@ -141,7 +144,10 @@ type OpinionatedReconciler struct {
 	client     PatchClient
 }
 
-const opinionatedReconcilerPatchStateKey = "grafana-app-sdk-opinionated-reconciler-create-patch-status"
+const (
+	opinionatedReconcilerPatchAddStateKey    = "grafana-app-sdk-opinionated-reconciler-create-patch-status"
+	opinionatedReconcilerPatchRemoveStateKey = "grafana-app-sdk-opinionated-reconciler-delete-patch-status"
+)
 
 // Reconcile consumes a ReconcileRequest and passes it off to the underlying ReconcileFunc, using the following criteria to modify or drop the request if needed:
 //   - If the action is a Create, and the OpinionatedReconciler's finalizer is in the finalizer list, update the action to a Resync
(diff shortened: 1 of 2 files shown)


charandas:
>As a side effect, actual delete events are no long propagated to the downstream.

This should be ok, right - since downstream is the one effectively conducting the delete operation and would be synchronously informed of that operation?

IfSentient:
> > As a side effect, actual delete events are no long propagated to the downstream.
> 
> This should be ok, right - since downstream is the one effectively conducting the delete operation and would be synchronously informed of that operation?

Yup, this is just a note on the behavior--we translate the "update" event that adds the `deletionTimestamp` into a downstream "delete" event, and swallow the real "delete" event, as the downstream reconciler has already processed our fake "delete" in order for us to remove the finalizer (and trigger the real "delete").

radiohead (approved):

Issue #515, opened 2024-11-28
Add GroupVersion to generated code


toddtreece:
requested by @kalleep 

> Would it be possible to generate constants for this that is shared? This part now looks a bit strange if you have several resources under the same GroupVersion.

[Original Comment](https://github.com/grafana/grafana/pull/97161#discussion_r1862204834)

PR #563, opened 2025-01-06
[codegen] Generate exported constants for group and version.

IfSentient:
Generate exported `Group`, `Version`, and `GroupVersion` constants for each generated kind's package (when grouping kinds by version, only one set of constants will be generated for the package). This allows for a more convenient way to reference these constants vs. using the Kind accessor method, and then getting its Group/Version, especially when kinds are grouped by version, so multiple kinds are in the same package, sharing the same constants.

This is implemented as a new generated file because it has to be generated only once per package, unlike the other kind files which exist once per kind.

Resolves https://github.com/grafana/grafana-app-sdk/issues/515

### DIFF
--- a/codegen/cuekind/generators_test.go
+++ b/codegen/cuekind/generators_test.go
@@ -60,8 +60,8 @@ func TestResourceGenerator(t *testing.T) {
 		files, err := ResourceGenerator(false).Generate(kinds...)
 		require.Nil(t, err)
 		// Check number of files generated
-		// 12 (6 -> object, spec, metadata, status, schema, codec) * 2 versions
-		assert.Len(t, files, 12)
+		// 14 (6 -> object, spec, metadata, status, schema, codec, constants) * 2 versions
+		assert.Len(t, files, 14, "should be 14 files generated, got %d", len(files))
 		// Check content against the golden files
 		compareToGolden(t, files, "go/groupbykind")
 	})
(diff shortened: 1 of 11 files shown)


codegen/cuekind/generators_test.go
```diff
@@ -60,8 +60,8 @@ func TestResourceGenerator(t *testing.T) {
 		files, err := ResourceGenerator(false).Generate(kinds...)
 		require.Nil(t, err)
 		// Check number of files generated
-		// 12 (6 -> object, spec, metadata, status, schema, codec) * 2 versions
-		assert.Len(t, files, 12)
+		// 14 (6 -> object, spec, metadata, status, schema, codec, constants) * 2 versions
```
charandas, line 63:
```suggestion
		// 14 (7 -> object, spec, metadata, status, schema, codec, constants) * 2 versions
```

codegen/cuekind/generators_test.go
```diff
@@ -70,8 +70,8 @@ func TestResourceGenerator(t *testing.T) {
 		files, err := ResourceGenerator(true).Generate(kinds...)
 		require.Nil(t, err)
 		// Check number of files generated
-		// 12 (6 -> object, spec, metadata, status, schema, codec) * 2 versions
-		assert.Len(t, files, 12)
+		// 14 (6 -> object, spec, metadata, status, schema, codec, constants) * 2 versions
```
charandas, line 73:
```suggestion
		// 14 (7 -> object, spec, metadata, status, schema, codec, constants) * 2 versions
```

charandas (approved):

radiohead (approved):

PR #565, opened 2025-01-06
Deprecate `simple.NewOperator` and update examples

IfSentient:
Added deprecation comment to `simple.NewOperator` and `simple.Operator`. `examples/operator/simple` was still using `simple.Operator`, so the code has been updated to use `simple.App` and `operator.Runner`. `examples/operator/opinionated` was building an operator from scratch, which is no longer advised, so the example has been removed.


examples/operator/simple/watcher/main.go
```diff
@@ -54,20 +69,37 @@ func main() {
...
+}
+
+type BasicModel struct {
+	Number int    `json:"numField"`
+	String string `json:"stringField"`
+}
+
+func NewApp(config app.Config) (app.App, error) {
 	// Set up the watcher
 	watcher := &simple.Watcher{
```
toddtreece, line 104: @radiohead mentioned that a reconciler was preferred over a watcher. if so, should the watcher be switched for a reconciler as well?

IfSentient: We have an example for each right now, I don't know if we want to remove the watcher examples completely yet (especially since codegen still builds a watcher by default at the moment)

toddtreece (approved):

PR #573, opened 2025-01-08
[CLI] Update/Remove/Rename CLI flags

IfSentient:
Clean up CLI flags by renaming some and removing others. Some functionality is also changed.

## Flag Changes
* `--selectors|-s` has been removed, and partially replaced by the `--manifest` flag. `--manifest` allows you to specify a selector for the app manifest, but does not allow for multiple selectors like `--selectors` did. This change was made as we no longer source kinds, but a manifest, and there should be one manifest per app.
* `--cuepath|-c` has been renamed to `--source|-s`, to better reflect the future goal of having different source formats vi the `--format` flag.
* `--crdencoding` has been renamed to `--defencoding`, as it now encodes both CRDs and app manifests
* `--crdpath` has been renamed to `--defpath`, as it dictates the location of more than just CRDs now (also app manifest)
* `--kindgrouping` has been renamed to `--grouping` to simplify (though I'm not 100% sure this is a positive change, perhaps just a neutral one to make the flag name shorter)

## Generated Code Changes

The `/resource` component of the generated code when grouping by kinds has been removed. To preserve this behavior in an existing project where you don't want the path to change, use `-g=pkg/generated/resource` (instead of the default `pkg/generated`).

Drive-by fix to generated operator and backend plugin code imports when grouping by group.

Local dev environment now also respects the `--manifest` flag (it didn't respect `--selectors` before).

This PR also contains a drive-by fix to `app.MultiRunner` for writing to a closed channel.


app/runner.go
```diff
@@ -56,7 +56,11 @@ func (m *MultiRunner) Run(ctx context.Context) error {
 			err := r.Run(propagatedContext)
 			wg.Done()
 			if err != nil && !timedOut {
-				errs <- err
+				select {
```
IfSentient, line 59: THis is a drive-by fix for an edge case that crops up every so often in tests where the channel is already closed because the runner has exited, but the goroutine hadn't finished yet.

cmd/grafana-app-sdk/generate.go
```diff
@@ -233,7 +203,7 @@ func generateKindsCue(modFS fs.FS, cfg kindGenConfig, selectors ...string) (code
 	// Manifest
 	var manifestFiles codejen.Files
 	var goManifestFiles codejen.Files
-	if cfg.GenerateManifest {
+	if cfg.CRDEncoding != "none" {
 		if cfg.CRDEncoding != "none" {
```
radiohead, line 207: Coulde we simplify this?

cmd/grafana-app-sdk/project.go
```diff
@@ -411,53 +397,53 @@ func projectAddComponent(cmd *cobra.Command, args []string) error {
...
 	}
 
+	manifests, err := manifestParser.Parse(os.DirFS(sourcePath), selector)
+	if err != nil {
+		return fmt.Errorf("error parsing manifest '%s': %v", sourcePath, err)
+	}
+	if len(manifests) == 0 {
+		return fmt.Errorf("no manifest found in '%s'", sourcePath)
+	}
+	manifest := manifests[0]
```
radiohead, line 438: Any chance we get `len(manifests) > 1`?

cmd/grafana-app-sdk/project.go
```diff
@@ -555,7 +541,7 @@ func addComponentOperator[G anyGenerator](projectRootPath string, generator G, s
 func addComponentBackend[G anyGenerator](projectRootPath string, generator G, selectors []string, pluginID string, groupKinds bool) error {
 	// Check plugin ID
 	if pluginID == "" {
-		return fmt.Errorf("plugin-id is required")
+		return fmt.Errorf("manifest group")
```
radiohead, line 544:
```suggestion
		return fmt.Errorf("manifest group is required")
```

cmd/grafana-app-sdk/project.go
```diff
@@ -555,7 +541,7 @@ func addComponentOperator[G anyGenerator](projectRootPath string, generator G, s
 func addComponentBackend[G anyGenerator](projectRootPath string, generator G, selectors []string, pluginID string, groupKinds bool) error {
```
radiohead, line 541: NIT, but not required:
```suggestion
func addComponentBackend[G anyGenerator](projectRootPath string, generator G, selectors []string, manifestGroup string, groupKinds bool) error {
```

cmd/grafana-app-sdk/project.go
IfSentient: There's not a normal situation where this should happen (we only allow one selector, and the parser should only return one), if we do, taking the first one should be fine, unless we'd rather error.

radiohead: I think we could error (for future-proofing, since explicit failure is easier to debug than weird behaviour) but I don't think it's a blocker – feel free to leave it as-is.

IfSentient:
I'll merge this in after today's release

PR #574, opened 2025-01-09
[apimachinery] Move pkg from grafana/grafana repo

Proximyst:
This moves in the `pkg/apimachinery` package from the Grafana repository: https://github.com/grafana/grafana/tree/9e8c1acd005b8ad0a8cc025cd633ef697047e1dd/pkg/apimachinery

It's recommended to review this commit-wise, as the first commit is a simple `cp` followed by a search-and-replace for the package imports. The later commits change stuff around.

This is still a WIP!


apimachinery/go.mod
```diff
@@ -0,0 +1,53 @@
+module github.com/grafana/grafana-app-sdk/apimachinery
+
+go 1.23.4
+
+require (
+	github.com/grafana/authlib v0.0.0-20250108202437-7a039176d884
```
ryantxu, line 6: dooh -- not sure when this full lib got added vs just the interfaces!

ryantxu:
Lets start with just logging and maybe the meta utility?  I looks like that mini package grew and grew while not looking 😬

Proximyst:
Sure thing! Do you think we should still call it apimachinery in that case, or should we perhaps have multiple subpackages? 🤔 

(Multiple subpackages or calling it logging would make the job of fixing [these](https://github.com/search?q=%22github.com%2Fgrafana%2Fgrafana-app-sdk%2Flogging%22&type=code) _much_ easier)

ryantxu:
@radiohead -- thoughts?

+1 for just a logging package -- zero dependencies :)

IfSentient:
@ryantxu @Proximyst +1 to just moving over the meta util to start, just to make the PR more digestible and easier to see if we have redundencies/things we might want to alter in either the code being moved or the existing app-sdk code.

Is there is a reason to move the `logging` package to be `apimachinery/logging`? The app-sdk logger is intended to be a top-level interface/package because it's used by the app-sdk and apps themselves, not specific to apimachinery (as we try to emphasize the idea that "apps are portable," so they aren't tied to being an API server or an operator or a plugin, they just can be run as one).

radiohead:
@ryantxu @Proximyst Same as Austin, +1 to just meta utils & logging. Don't have strong preferences over naming.

radiohead:
@Proximyst @ryantxu @IfSentient should we close this PR for now? I think the best way forward would be to move the code from apimachinery piecemeal, like we did with logging?

ryantxu:
+1 to close it

radiohead:
Alright, let's close it for now!

Issue #236, opened 2024-03-28
Adopt Release Branch Strategy


IfSentient:
Currently, the grafana-app-sdk releases from `main` and has commit-based tags. This prevents us from easily offering continued support for older releases. We should investigate moving toward release branches for each minor release (while still pre-1.0). To do so, we will need to ensure the following:
- [ ] The release script(s) work with release branches instead of main (either cutting a new one for a minor version bump, or using an existing one for a patch)
- [ ] PR's into release branches follow the same rules as PRs into `main`
- [ ] Dependabot can cut dependency upgrade PRs into active release branches as well as `main` (otherwise backporting dependency upgrades becomes higher-toil work)
- [ ] The release process needs better documentation

PR #575, opened 2025-01-09
[logging] Make a proper submodule

Proximyst:
This makes logging a proper submodule, such that depending on the entirety of the SDK is not necessary just to get to work with logging.

Split off from #574.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -142,6 +142,7 @@ jobs:
           grafana-app-sdk project init "test-project"
           rm -rf kinds/manifest.cue
           printf "\nreplace github.com/grafana/grafana-app-sdk => $(readlink -f ../grafana-app-sdk)\n" >> go.mod
+          printf "\nreplace github.com/grafana/grafana-app-sdk/logging => $(readlink -f ../grafana-app-sdk/logging)\n" >> go.mod
           printf "\nreplace github.com/grafana/grafana-app-sdk/plugin => $(readlink -f ../grafana-app-sdk/plugin)\n" >> go.mod
           cp ../grafana-app-sdk/codegen/cuekind/testing/customkind.cue kinds/customkind.cue
           grafana-app-sdk generate --selectors=customManifest
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ all: check-go-version deps lint test build
 
 deps: $(GOSUM) $(GOWORKSUM)
 $(GOSUM): $(SOURCES) $(GOMOD)
-	$(foreach mod, $(dir $(GOMOD)), cd $(mod) && go mod tidy -v;)
+	$(foreach mod, $(dir $(GOMOD)), pushd . && cd $(mod) && go mod tidy -v; popd;)
 
 $(GOWORKSUM): $(GOWORK) $(GOMOD)
 	go work sync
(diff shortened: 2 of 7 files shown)


IfSentient:
@Proximyst I think you'll need to add a replace directive [here](https://github.com/grafana/grafana-app-sdk/blob/main/.github/workflows/test.yml#L145) in the GH test action as well for integration tests to succeed.

Proximyst:
@IfSentient Thanks, I was thinking of how we might make that work out!

Do you think we should remove the replace directives afterwards, or shall we just keep them replaced (as they're the same project anyhow)?

IfSentient:
I think we'll need a new release process for this submodule--it should very rarely, if ever, change, so I'm of the opinion that its release versions are decoupled from the main and plugin module release versions. I'm also of this opinion because trying to couple it becomes complicated due to the face that the main module depends on it (unlike plugin, which main does not depend on), so we'd have to "release" (tag) the logging module, then add a commit to main using the latest before the release of that module, which becomes messy. Though this probably brings up how important it is to get https://github.com/grafana/grafana-app-sdk/issues/236 implemented, as it may make these things simpler.

Thoughts on this @radiohead @ryantxu @toddtreece @Proximyst ?

IfSentient:
> @IfSentient Thanks, I was thinking of how we might make that work out!
> 
> Do you think we should remove the replace directives afterwards, or shall we just keep them replaced (as they're the same project anyhow)?

We'll have to ditch the replace directive, because having the replace breaks `go install`, so users won't be able to install the CLI that way if we have it.

Proximyst:
I think it would be fine to decouple its version from the other modules, though I do wonder if that's worth the time at all. Ergonomically, I'd much rather just use the same version as the other modules do, and let it be unchanged for many releases after one another.

I can see the messiness happening upon changes, indeed. Though I don't imagine it would be a significant feat to split logging changes out while retaining backwards compatibility for when we to do a change. I'd guesstimate that would be optimising for a workflow that happens once a quarter (at most!).

radiohead:
> I think we'll need a new release process for this submodule--it should very rarely, if ever, change, so I'm of the opinion that its release versions are decoupled from the main and plugin module release versions. I'm also of this opinion because trying to couple it becomes complicated due to the face that the main module depends on it (unlike plugin, which main does not depend on), so we'd have to "release" (tag) the logging module, then add a commit to main using the latest before the release of that module, which becomes messy. Though this probably brings up how important it is to get #236 implemented, as it may make these things simpler.
> 
> Thoughts on this @radiohead @ryantxu @toddtreece @Proximyst ?

I think _ideally_ we should decouple it, but for now I'd just leave it as-is, since it doesn't cause any blocking issues, IIUC.

MissingRoberto (approved):
lgtm ❤️

radiohead (approved):

.github/workflows/test.yml
```diff
@@ -142,6 +142,7 @@ jobs:
           grafana-app-sdk project init "test-project"
           rm -rf kinds/manifest.cue
           printf "\nreplace github.com/grafana/grafana-app-sdk => $(readlink -f ../grafana-app-sdk)\n" >> go.mod
+          printf "\nreplace github.com/grafana/grafana-app-sdk/logging => $(readlink -f ../grafana-app-sdk/logging)\n" >> go.mod
```
macabu, line 145: Do we need this line when there's a replace of the logging in the gomod?

macabu (approved):
Just a question but non blocking

.github/workflows/test.yml
Proximyst: Yes, unfortunately. We will remove it once merged, however!

macabu: Sorry, I completely skipped over this was already discussed in the thread 😶‍🌫️

PR #605, opened 2025-01-22
Added resource.WrappedObject for creating a resource.Object-implmenting type from a non-spec-based type

IfSentient:
Added `resource.WrappedObject` for creating a `resource.Object`-implmenting type from a kubernetes type which doesn't use 'spec' in its schema (for example: ClusterRole). Also added `resource.PassthroughJSONCodec` for said type which does no additional logic aside from JSON Marshal/Unmarshal (unlike `resource.JSONCodec` which uses the accessors in `resource.Object` to structure the resulting JSON).

### DIFF
--- a/resource/wrappedobject.go
+++ b/resource/wrappedobject.go
@@ -1,0 +1,320 @@
+package resource
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


resource/wrappedobject.go
```diff
@@ -0,0 +1,320 @@
...
+	updt := time.Time{}
+	createdBy := ""
+	updatedBy := ""
+	annotations := o.Object.GetAnnotations()
+	if annotations != nil {
+		strUpdt, ok := annotations[AnnotationUpdateTimestamp]
+		if ok {
+			updt, err = time.Parse(time.RFC3339, strUpdt)
+			if err != nil {
+				// HMMMM
```
onprem, line 262: We can try another format as a fallback here if there's any other probable format, but I guess in the end we have to ignore the error.

onprem (approved):
LGTM!

PR #613, opened 2025-01-28
Update APIVersion in AppManifest Jenny to Match AppManifest CRD Definition

IfSentient:
Update the `apiVersion` in the generated AppManifest CR to match the `apiVersion` in the [CUE definition of the AppManifest CRD](https://github.com/grafana/grafana-app-sdk/blob/main/app/manifest.cue#L14).


charandas (approved):
LGTM!

onprem (approved):

PR #614, opened 2025-01-28
Replace the outdated 'Writing an Operator' doc with two new docs

IfSentient:
Replace the outdated 'Writing an Operator' doc with two new docs: 'Writing an App' and 'Writing a Reconciler', which contain similar information and further expand on the app concept. Update docs with references to the old 'Writing an Operator' doc.


docs/operators.md
```diff
@@ -54,7 +49,11 @@ package main
 import (
 	"context"
 	"fmt"
-	
+	"os"
```
IfSentient, line 52: The code updates here are a drive-by as the example still used `simple.NewOperator`. This whole doc may also need some more updates--while it is factually correct, it may not be a good first-brush resource for the operator pattern, as it talks about the `operator` package construction of an operator, which we now hide behind `simple.App`. I added a quick into paragraph to add that context because I'm not sure if/when/how much we want to update this doc.

docs/writing-a-reconciler.md
```diff
@@ -0,0 +1,164 @@
+# Writing a Reconciler (or Watcher)
+
+A Reconciler is a logical pattern which takes the state of a set of objects in the API server, and propagates changes to other components based on that state. 
+This logic is asnchronous, and works by having a function which is called when a change occurs in the set of resources the reconciler is watching, 
+which provides the current state of the object, and the type of change that occurred (Create, Update, Delete). 
+For more information, see the [Asynchronous Business Logic section in Platform Concepts](application-design/platform-concepts.md#asynchronous-business-logic).
+
+## Reconciler vs Watcher
```
IfSentient, line 8: This section (and the considerations section) are taken mostly unchanged from the old Writing an Operator doc, with small updates where needed.

docs/writing-a-reconciler.md
```diff
@@ -0,0 +1,164 @@
...
+* The reconciler is taking action on _every_ consumed event. Finding ways to escape from a reconcile or watcher event early will help your overall program logic.
+* All objects for the kind(s) you are watching are cached to memory by default. This can be customized by using a different informer implementation, such as `operator.CustomCacheInformer`. Custom informers can be used in `simple.App` with `AppConfig.InformerConfig.InformerSupplier`, or by using your own custom `app.App` implementation.
```
IfSentient, line 25: `AppConfig.InformerConfig.InformerSupplier` is being added in [this PR](https://github.com/grafana/grafana-app-sdk/pull/612), and is not yet merged at time of writing.

onprem (approved):
LGTM! I only have one minuscule nit.

radiohead (approved):

PR #619, opened 2025-01-30
Small fixes for the Issue Tracker tutorial

pstibrany:
This PR updates the Issue Tracker tutorial with few small fixes.

Apart from these fixes, I had to change `issue-tracker-project-app` part of URL (eg. in `curl http://grafana.k3d.localhost:9999/api/plugins/issue-tracker-project-app/resources/v1/issues` command) to `issuetrackerprojectapp`. I'm not sure why, or whether this is specific to my installation somehow, or version, so I didn't update the commands in this PR.

### DIFF
--- a/docs/tutorials/issue-tracker/04-boilerplate.md
+++ b/docs/tutorials/issue-tracker/04-boilerplate.md
@@ -353,20 +353,20 @@ That leaves us with just our varying TypeScript files.
 
 ```TypeScript
 function PageOne() {
-    const s = useStyles2(getStyles);
+  const s = useStyles2(getStyles);
 
-    return (
-        <PluginPage>
-            <div data-testid={testIds.pageOne.container}>
-            This is page one.
-    <div className={s.marginTop}>
-    <LinkButton data-testid={testIds.pageOne.navigateToFour} href={prefixRoute(ROUTES.Four)}>
-    Full-width page example
-    </LinkButton>
-    </div>
-    </div>
+  return (
+    <PluginPage>
+      <div data-testid={testIds.pageOne.container}>
+        This is page one.
+        <div className={s.marginTop}>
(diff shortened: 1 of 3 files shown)


docs/tutorials/issue-tracker/04-boilerplate.md
```diff
@@ -254,35 +263,18 @@ In our list handler boilerplate, we can see we grab filters from the query, if p
...
-}
+app.Manage(pluginID, newInstanceFactory(logger), app.ManageOpts{
+	TracingOpts: tracing.Opts{
+		CustomAttributes: []attribute.KeyValue{
+			attribute.String("plugin.id", pluginID),
+		},
+	},
+})
 ```
-The important thing to look at is the `kubeconfig.CachingInitializer` being used for the service initializer func. This is another SDK library which allows us to define an initializer for a service which will be called only once per unique kube config. We'll get more in-depth on what this is and why we need to do this when we begin writing our back-end code, but I want to point this out.
```
pstibrany, line 283: I've removed this section, as `kubeconfig.CachingInitializer` is not used anymore in the generated code, and entire `main` was mostly replaced by call to `app.Manage`.

IfSentient:
> Apart from these fixes, I had to change `issue-tracker-project-app` part of URL (eg. in `curl http://grafana.k3d.localhost:9999/api/plugins/issue-tracker-project-app/resources/v1/issues` command) to `issuetrackerprojectapp`

Ah, I think I know why, and it should be fixed in the docs, let me double-check on my local as well. The `group` auto-generates from the `appName` in your manifest, and removes non-alphanumeric characters. I fixed it previously in the `curl` commands and in the YAML for creating a new issue, but clearly I missed it in the URL links.

IfSentient (approved):
Thanks for these updates/fixes!

PR #624, opened 2025-02-03
Bump grafana/cog to v0.0.21

K-Phoen:
This version of cog comes with a better support for "out-of-root" objects in CUE.

Relates to https://github.com/grafana/grafana/pull/99966


IfSentient (changes requested):
Looks like some of the codegen output changed, can you run
```
make regenerate-codegen-test-files && rm codegen/testing/golden_generated/go/groupbykind/customapp_manifest.go.txt
```
To get the updated codegen outputs added? The `rm` is there as there is a small bug that also creates an extra file that should be removed before committing

K-Phoen:
Done :)

The changes should only be related to the order of some declarations (which is directly linked to the fix in cog)

IfSentient (approved):
LGTM

PR #626, opened 2025-02-04
Add operator health endpoints and share mux with metrics

charandas:
### Introduction

Adds functionality that allows app developers to specify their own implementation of health checks. The inspiration for this came from `metrics.Provider`, which is an interface. An informer, watcher or controller is able to implement this interface and the runners automatically expose metrics from the said components. Similarly, with this PR, we introduce two Interfaces in the brand new `health` pacakge that app developers can optionally implement their own checks using the `Checker` interface, each checks implementation being an implementation of `Check` interface.

Additionally, the `health` package provides the `Observer` implementation. Observer runs the aforementioned checks periodically and a cached value can be fetched at any time. This is to prevent a DDoS attack in case an external actor exercises the status endpoints ad-infinitum.

#### Details

* The checks implementation is merged with the server mux that already exists in the metrics server. Some newer config structs are introduced to streamline providing the port and interval configuration. `9090` being the default for backwards compat when such config is omitted.
* `operator/runner` and `app/runner` are updated to provide this functionality.
* `simple` package (included the deprecated operator) components are also updated.
* `examples` were minimally updated as well to pass in the new config.


operator/operator_server.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+type OperatorServer struct {
+	Mux  *http.ServeMux
```
charandas, line 15: Tried keeping the Mux private as well as OperatorServer private to the go package, but that didn't fly so well when using it with the now deprecated `simple/operator.go`.

operator/operator_server.go
```diff
@@ -0,0 +1,64 @@
...
+	Port int
+}
+
+func (s *OperatorServer) RegisterLivenessHandler(handler func(http.ResponseWriter) error) error {
+	if handler == nil {
+		return errors.New("liveness handler cannot be nil")
+	}
+	s.Mux.Handle("/livez", http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		_ = handler(w)
```
charandas, line 24: Is error suppression here when calling the custom handler something that should be accompanied with a warning if it errors?

operator/operator_server.go
charandas: Could go back to private if we can find another way to keep the deprecated code building.

metrics/config.go
```diff
@@ -6,7 +6,6 @@ import "github.com/prometheus/client_golang/prometheus"
 type ExporterConfig struct {
 	Registerer prometheus.Registerer
 	Gatherer   prometheus.Gatherer
-	Port       int
```
charandas, line 9: Will this be considered a breaking change?

metrics/metrics.go
```diff
@@ -60,30 +59,8 @@ func (e *Exporter) RegisterCollectors(metrics ...prometheus.Collector) error {
 	return nil
 }
 
-// Run creates an HTTP server which exposes a /metrics endpoint on the configured port (if <=0, uses the default 9090)
-func (e *Exporter) Run(stopCh <-chan struct{}) error {
-	mux := http.NewServeMux()
-	mux.Handle("/metrics", promhttp.InstrumentMetricHandler(
+func (e *Exporter) RegisterMetricsHandler() {
```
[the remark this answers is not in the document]
IfSentient: Another solution (in fact, I think it would be useful to have both this method and the following) would be to export the handle function with something like `func (e *Exporter) HTTPHandler() http.Handler`

IfSentient (commented):
@radiohead made a very good suggestion when I was talking with him just now, which is we could model the readiness health check with an interface like [metrics.Provider](https://github.com/grafana/grafana-app-sdk/blob/main/metrics/metrics.go#L40) and have a healthchecker that checks every Runnable for conformance to that interface and adds those healthchecks to the overall readiness check. Something like
```go
type HealthCheck interface {
    HealthCheck(context.Context) error
}
```
And then have a HealthChecker that allows you to register HealthChecks (just like the metrics.Exporter `RegisterCollectors`).

In fact, I think modeling a `HealthChecker` basically the same way the `metrics.Exporter` is probably a good way forward for this, it still allows for customization of the health checks (just implement `HealthCheck` to supply your own readiness check), and becomes extensible for other runners to use. I think this should probably belong it its own package, and then be used by `operator.Runner`. Something like
```go
type HealthChecker struct {
    // Unexported fields
}

// RegisterHealthCheck adds a HealthCheck to be run by the HealthChecker
func (*HealthChecker) RegisterHealthCheck(HealthCheck) {}

// RunHealthCheck runs all registered health checks and returns any errors combined into a single error
func (*HealthChecker) RunHealthCheck(context.Context) error {}

// HTTPHandler returns an http.Handler that performs the health checks
func (*HealthChecker) HTTPHandler() http.Handler {}
```
Then we can use this in the mux for the readiness check (liveness check should return ok if the server is up), alongside the metrics handler.

charandas:
>Then we can use this in the mux for the readiness check (liveness check should return ok if the server is up), alongside the metrics handler.

Does the health check interface then only apply to readiness check? And liveness check doesn't require any customization, so just mount a default behavior to the mux without offering any overrides for it?

IfSentient:
> > Then we can use this in the mux for the readiness check (liveness check should return ok if the server is up), alongside the metrics handler.
> 
> Does the health check interface then only apply to readiness check? And liveness check doesn't require any customization, so just mount a default behavior to the mux without offering any overrides for it?

Yeah, I think the take-away for me was that there's no reason to allow/need liveness customization, as it should return ok if the HTTP server is running (making it live).

metrics/metrics.go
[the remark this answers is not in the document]
IfSentient: I think exporting the handler is probably the most extensible way of handling it, it allows the owner of the mux to decide under what HTTP endpoint the handler resides. Adding the "pass the mux" function would mainly be a way to have the default `/metrics` endpoint used, but you're right that it's probably overkill to also include that.

health/health.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+var _ HealthCheck = (*HealthChecker)(nil)
+
+// RegisterHealthCheck adds a HealthCheck to be run by the HealthChecker
+func (c *HealthChecker) RegisterHealthCheck(fn HealthCheckFunc) {
+	c.fns = append(c.fns, fn)
+}
+
+// RunHealthCheck runs all registered health checks and returns any errors combined into a single error
```
charandas, line 26: Guess it doesn't currently do that, need to improve to accumulate errors.

operator/runner.go
```diff
@@ -49,17 +52,22 @@ func NewRunner(cfg RunnerConfig) (*Runner, error) {
 		cfg.Port = 9090
 	}
 
+	hc := cfg.HealthCheck
+	if isInterfaceNil(hc) {
```
charandas, line 56: not sure if there is a util elsewhere I can use for this, but go is pretty bad at checking an interface value for nil.

health/health.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// HTTPHandler returns an http.Handler that performs the health checks
+func (c *HealthChecker) HTTPHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if c.runHealthChecks(r.Context()) != nil {
+			w.WriteHeader(200)
+			_, _ = w.Write([]byte("ok"))
+			return
+		}
```
radiohead, line 45: Can we make this async? i.e. periodically run the HCs in a separate goroutine and just load the result here, instead of running them on demand. The issue with the sync approach is that a misconfigured HC period can result in a DoS-like attack on the service (especially if HC is expensive, like pinging a DB connections).

health/health.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+	"net/http"
+)
+
+type HealthCheck interface {
+	RegisterHealthCheck(fn HealthCheckFunc)
+	HTTPHandler() http.HandlerFunc
+}
+
+type HealthCheckFunc = func(context.Context) error
```
IfSentient, line 14: Instead of a function, can we make this an interface which has this function? That way it can behave like `metrics.Provider` and we can auto-add health checks from any component which implements the interface when we start a runner (like we do [here](https://github.com/grafana/grafana-app-sdk/blob/main/operator/runner.go#L218) with metrics), we'd just need to have the various runnables in `app` also implement the health check interface and call the same methods downstream like [with metrics.Provider](https://github.com/grafana/grafana-app-sdk/blob/main/app/runner.go#L107)

charandas: There is already an interface like this at the moment:

```
type HealthCheck interface {
	RegisterHealthCheck(fn HealthCheckFunc)
	HTTPHandler() http.HandlerFunc
}

type HealthCheckFunc = func(context.Context) error

type HealthChecker struct {
	fns []HealthCheckFunc
}
```

`HealthCheckFunc` is syntactic sugar which allows me to type less when implementing the struct. Is that okay? If the HealthCheck interface is going to expose an HTTPHandler anyway, I see very little point to exposing the function that the same handler invokes. Maybe I am missing something?

operator/runner.go
charandas: `nil` check will fail if one passes an empty interface value inadvertently - which will be a runtime memory exception when invoking methods on that interface.

To pass an empty interface value, all you got to do in Go is:

```go
var someVal HealthCheck
passThisToSomeFuncWhichOnlyChecksNil(someVal) // crashes
```

health/health.go
charandas: What you say, makes sense! My only last confusion is this looks like inception in your previous comment:

```
func (*HealthChecker) RegisterHealthCheck(HealthCheck) {}
```

The interface can certainly be close to what you said:

```go
type HealthCheck interface {
	RegisterHealthCheck(fn HealthCheckFunc) // where HealthCheckFunc is also func (context.Context) error
	HTTPHandler() http.HandlerFunc
	RunHealthCheck(context.Context) error
}
```

But I was under the impression that HealthChecker was implementing HealthCheck, hence me coding the RegisterHealthCheck using a func parameter so as to avoid circularity here. Am I misunderstanding this?

charandas: >we'd just need to have the various runnables in app also implement the health check interface

That part again makes me think, I am getting the interface/struct relationship wrong.

The HTTPHandler part also makes it confusing. If all the various runnables implement their own HTTPHandler, who will fuse their result? I think fusing the results happens in lieu of a single HTTPHandler.

IfSentient: In this case, we'd want something like the HealthChecker to be a struct with a function that allows you to register a health check, like you have with the `HealthCheckFunc`, but instead of providing a function, it would accept an interface that any component can implement, and then the runner can check all of the components it has to see if they implement the HealthCheck interface. If they do, it can pass them to RegisterHealthCheck to get their health checks added. This allows for an app author to add runnables which may have health checks on them, and get them automatically picked up by the runner and added to the HealthChecker, the same way an app author can implement `metrics.Provider` on their arbitrary runnable and have the metrics get gathered in the metrics server.

If we wanted to, we could even name this similarly to the metrics version, `health.Provider`:
```go
type Provider interface {
    HealthCheck(context.Context) error // this is the same as HealthCheckFunc, just turned into an interface
}

// other code
func (*HealthChecker) RegisterHealthCheck(provider Provider)
```
Then we can implement `health.Provider` in the runnables the same way [we do for metrics.Provider](https://github.com/grafana/grafana-app-sdk/blob/main/app/runner.go#L107-L115), and be able to use that to register all health checks for all runnables provided to the runner (given that's how we do it for [metrics](https://github.com/grafana/grafana-app-sdk/blob/main/operator/runner.go#L218-L221)) (and to get the InformerController and its constituent informers to expose health checks, we just implement `health.Provider` in `InformerController` for it to get automatically picked up).

charandas: I added a runner because the HealthCheck interface's RunHealthCheck needs to be repeated on an interval. On a second thought, the runner itself could be a part of the interface/struct holding all of this.

The decision about that hinges on whether the HTTPHandler will be centralized on the main interface itself, as well. If yes, then I don't an issue with combining the runner into it.

operator/runner.go
```diff
@@ -26,12 +29,16 @@ import (
...
-	metricsServer *metricsServerRunner
-	startMux      sync.Mutex
-	running       bool
-	runningWG     sync.WaitGroup
+	config            RunnerConfig
+	webhookServer     *webhookServerRunner
+	metricsExporter   *metrics.Exporter
+	serverRunner      *app.SingletonRunner
+	healthCheck       health.HealthCheck
+	healthCheckRunner *health.HealthCheckRunner
```
radiohead, line 37: A couple of nits – these should IMHO be just one component, that implements `Run` and `AddChecks`. Also, the interface should be defined here (i.e. on the caller side), not on the implementation side, following the Go pattern of "return concrete types, use interface types".

Something like:
```go
type HealthCheck interface {
	app.Runnable
	AddChecks(...health.Check) error
}
```

health/health.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"sync"
+)
+
+type HealthCheck interface {
```
radiohead, line 11: NIT – I don't think we need to define this interface here, since we should be exporting concrete types.

operator/operator_server.go
```diff
@@ -0,0 +1,48 @@
...
+func (s *OperatorServer) RegisterMetricsHandler(handler http.Handler) {
+	s.Mux.Handle("/metrics", handler)
+}
+
+func (s *OperatorServer) registerHealthHandlers() error {
+	s.Mux.Handle("/livez", http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		_, _ = w.Write([]byte("ok"))
+		return
+	}))
+	s.Mux.Handle("/readyz", s.HealthCheck.HTTPHandler())
```
radiohead, line 30: NIT – since the server already knows about the existence of healthchecks, we don't need to delegate HTTP handler implementation to it, we can just define it here, same as liveness handler:
```suggestion
	s.Mux.Handle("/readyz", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if err := s.HealthCheck.Status(r.Context()); err != nil {
				_, _ = w.WriteHeader(http.StatusInternalServerError)
				// TODO: we can add the ability for HC to return detailed status, in which case we'd want to include it in the body.
				return
		}

		_, _ = w.Write([]byte("ok"))
		return
	}))
```

operator/runner.go
charandas: Of course! converting it to interface is definitely on my plate. Combining them makes it easier - which I had a [hint](https://github.com/grafana/grafana-app-sdk/pull/626#discussion_r1950005018) about doing, just wanted to confirm my thinking with you both.

charandas: >the Go pattern of "return concrete types, use interface types".

Learned something new!

charandas: Some projects I have worked with so far define the interface and struct in the same file: https://github.com/grafana/authlib/blob/bf4600a17569ead7f7ddd49ff00136916896c4ce/authn/verifier.go#L69-L78

Is that considered difficult on the eyes?

operator/operator_server.go
charandas: That sounds better for sure, I will need to adjust the interface a little to include `Status` first.

operator/runner.go
```diff
@@ -214,14 +243,28 @@ func (s *Runner) Run(ctx context.Context, provider app.Provider) error {
...
+		s.server.RegisterMetricsHandler(s.metricsExporter.HTTPHandler())
 	}
 
+	// Health
+	s.healthCheck.RegisterHealthCheck(func(ctx context.Context) error {
+		if !s.running {
+			return errors.New("app has not started yet")
+		}
+		return nil
+	})
```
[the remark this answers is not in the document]
charandas: I removed it for now, but just to be clear, I am trying to implement something for OperatorRunner and it doesn't use the MultiRunner - I also need access to the private `running` to implement the check. As such, to me, it would make sense to implement it inside OperatorRunner.

health/health.go
charandas: I changed the package to only define interfaces for now, because I am not sure, your other suggestions are wanting the implementation in this package. Maybe you do? If so, I just need more clarification.

For now, the serverRunner which I introduced in the operator runner is the only one place where I am fully implementing the interfaces and more. The more part is the actual running of checks on an interval. I haven't seen any comments about making an interface for that yet. So it's only on the concrete implementation.

operator/runner.go
charandas: Changed to a simple nil check for now.

operator/operator_server.go
```diff
@@ -0,0 +1,119 @@
...
+			_, _ = w.Write([]byte("readiness check failed: " + err.Error()))
+			return
+		}
+		_, _ = w.Write([]byte("ok"))
+		return
+	}))
+	return nil
+}
+
+func (s *OperatorServer) Run(ctx context.Context) error {
```
charandas, line 49: My async go is pretty novice. Please shred it with your focused eyes if need be.

charandas:
Currently, it's all really the boilerplate and the checks itself are empty. Until I know how to implement one simple check of using operator's `running`, I will hang tight. I think some of the recommendations read as circular and need more clarification.

TL; DR: I am still looking for a way to define the check itself - something I had already done for the operator runner but retracted based on some feedback above.

operator/runner.go
```diff
@@ -214,17 +238,27 @@ func (s *Runner) Run(ctx context.Context, provider app.Provider) error {
...
+	// Add Metrics+Health cumulative server runnable, healthcheck based on "running" should
+	// be passed down to the serverRunner
+	runner.AddRunnable(s.serverRunner)
+
 	return runner.Run(ctx)
 }
 
+// HealthChecks implements HealthChecker
+func (s *Runner) HealthChecks() []health.Check {
+	return []health.Check{s.healthCheck}
```
charandas, line 259: Currently, `{nil}`, needs to be something.

operator/runner.go
radiohead: > What if you have multiple call sites spread across various files?

You can define your imported interface in one place, I'd put it in the root package file (i.e. `package/package.go` or `package/types.go`). But in my experience this has rarely been the case.

> Some projects I have worked with so far define the interface and struct in the same file.

Returning and interface is usually done when you have a package that has multiple implementations of the same one. Still, personally I always to prefer to return concrete types from initializers and use interface compatibility checks.

https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a is a good read on why's and how's.

operator/runner.go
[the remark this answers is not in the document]
radiohead: If you need to add an extra HC for just the operator runner, you could append it to the list of HCs separately, would that work?

operator/operator_server.go
radiohead: Oh yeah, that was a typo, sorry 😅

health/health.go
radiohead: > your other suggestions are wanting the implementation in this package
Yup, that's correct!

> The more part is the actual running of checks on an interval.
That should IMHO be part of the implementation here.

charandas: here, meaning in the "health" package? just a struct that is reusable across the board?

radiohead: Yeah, sorry.

Something like:

```go
package health

import (
	"context"
	"errors"
	"sync"
	"time"
)

type Check interface {
	HealthCheck(context.Context) error
	HealthCheckName() string
}

type Checker interface {
	HealthChecks() []Check
}

type Observer struct {
	checks      []Check
	checkLock   sync.RWMutex
	runInterval time.Duration
	runLock     sync.RWMutex
	runErr      error
}

func NewObserver(interval time.Duration) *Observer {
	return &Observer{
		checks:      []Check{},
		runInterval: interval,
	}
}

func (c *Observer) AddChecks(checks ...Check) {
	c.checkLock.Lock()
	c.checks = append(c.checks, checks...)
	c.checkLock.Unlock()
}

func (c *Observer) Status() error {
	c.runLock.RLock()
	defer c.runLock.RUnlock()

	return c.runErr
}

func (c *Observer) Run(ctx context.Context) error {
	t := time.NewTicker(c.runInterval)
	defer t.Stop()

	var lastErr error
	for {
		select {
		case <-ctx.Done():
			return lastErr
		case <-t.C:
			lastErr = c.collectChecks(ctx)

			c.runLock.Lock()
			c.runErr = lastErr
			c.runLock.Unlock()
		}
	}
}

func (c *Observer) collectChecks(ctx context.Context) error {
	c.checkLock.RLock()
	defer c.checkLock.RUnlock()

	var errs error

	for _, check := range c.checks {
		if err := check.HealthCheck(ctx); err != nil {
			errs = errors.Join(errs, err)
		}
	}

	return errs
}
```

app/runner.go
```diff
@@ -39,6 +41,9 @@ type MultiRunner struct {
 	// before stopping execution and returning a timeout error instead of exiting gracefully.
 	// If ExitWait is nil, Run execution will always block until all Runners have exited.
 	ExitWait *time.Duration
+	// HelthCheckInterval is the duration at which the runner will periodically run the registered health checks. The
+	// cached result is what's returned by the ready endpoint.
+	HealthCheckInterval time.Duration
```
IfSentient, line 46: Since this isn't used in the MultiRunner, I think it's probably better to track it externally (in the caller of the health checks).

app/runner.go
```diff
@@ -114,6 +119,20 @@ func (m *MultiRunner) PrometheusCollectors() []prometheus.Collector {
 	return collectors
 }
 
+// HealthChecks implements HealthChecker
```
IfSentient, line 122: NIT: `implements health.Checker`

app/runner.go
```diff
@@ -114,6 +119,20 @@ func (m *MultiRunner) PrometheusCollectors() []prometheus.Collector {
...
+func (m *MultiRunner) HealthChecks() []health.Check {
+	// TODO: there isn't a running property on the multirunner itself, should there be?
+	checks := []health.Check{}
+
+	for _, runner := range m.Runners {
+		if cast, ok := runner.(health.Checker); ok {
+			checks = append(checks, cast.HealthChecks()...)
+		}
```
IfSentient, line 130: We should also check if `runner` implements `health.Check` (if it doesn't implement `heath.Checker`), and it it does, add it to `checks`

health/health.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+)
+
+type Check interface {
+	HealthCheck(context.Context) error
+	HealthCheckName() string
+}
+
+type Checker interface {
+	// HealthChecks runs all registered health checks and returns any errors combined into a single error
```
IfSentient, line 13:
```suggestion
	// HealthChecks returns a slice of all Check instances which should be run for health checks
```

operator/operator_server.go
```diff
@@ -0,0 +1,118 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/grafana/grafana-app-sdk/health"
+)
+
+type OperatorServer struct {
```
IfSentient, line 17: Perhaps this should be named something like `ObservabilityServer`/`MetricsServer`/`HealthServer`, as it's not really exclusive to being an operator (and doesn't contain the operator endpoints for webhooks)? 

We may even want to put it in a different package to be shared by other things, but that's probably a future consideration for now.

operator/operator_server.go
```diff
@@ -0,0 +1,118 @@
...
+	Checks              []health.Check
+	Mux                 *http.ServeMux
```
IfSentient, line 19: I think we should avoid exporting the mux if we can here, to avoid handlers from being overwritten.

operator/operator_server.go
```diff
@@ -0,0 +1,118 @@
...
+func (s *OperatorServer) runChecks(ctx context.Context) error {
+	s.healthCheckErrMu.Lock()
+	defer s.healthCheckErrMu.Unlock()
+
+	fmt.Println("Running health check...")
+
+	var allErrors error
+	for _, check := range s.Checks {
+		if err := check.HealthCheck(ctx); err != nil {
+			allErrors = errors.Join(allErrors, err)
```
IfSentient, line 111: Can we include the healthcheck name (`check.HealthCheckName()`) in the error we add here? Something like 
```go
allErrors = errors.Join(allErrors, fmt.Errorf("%s: %w", check.HealthCheckName(), err))
```

operator/runner.go
```diff
@@ -214,17 +238,27 @@ func (s *Runner) Run(ctx context.Context, provider app.Provider) error {
...
 			return err
 		}
-		runner.AddRunnable(s.metricsServer)
+
+		s.server.RegisterMetricsHandler(s.metricsExporter.HTTPHandler())
 	}
 
+	// Add Metrics+Health cumulative server runnable, healthcheck based on "running" should
+	// be passed down to the serverRunner
+	runner.AddRunnable(s.serverRunner)
```
IfSentient, line 252: I think here we also want to run through `runner.HealthChecks()` and register all of them with `s.serverRunner`.

IfSentient (changes requested):

health/health.go
charandas: Incorporated this! Thank you.

health/observer.go
```diff
@@ -0,0 +1,80 @@
...
+			errs = errors.Join(errs, err)
+		}
+	}
+
+	return errs
+}
+
+func (c *Observer) CollectInitialChecks(ctx context.Context) error {
+	var result error
+	// is this right? am I overthinking it with sync.Once?
```
charandas, line 75: question 1

health/observer.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+func (c *Observer) AddChecks(checks ...Check) {
+	c.checkLock.Lock()
+	c.checks = append(c.checks, checks...)
+	c.checkLock.Unlock()
+}
+
+func (c *Observer) Status() error {
```
IfSentient, line 32: Another possibility here would be to have a `CheckResult` object with `Name` and `Error`, which would be `nil` for successful checks, so a caller could report all checks in their output.

IfSentient: We could have the `CheckResult` look something like this:
```
type CheckResult struct {
    Name string
    Error error
}

func (c CheckResult) String() {
    status := "ok"
    if c.Error != nil {
        status = c.Error.String()
    }
    return fmt.Sprintf("%s: %s", c.Name, status)
}
```

health/observer.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (c *Observer) collectChecks(ctx context.Context) error {
+	c.checkLock.RLock()
+	defer c.checkLock.RUnlock()
+
+	var errs error
+
+	for _, check := range c.checks {
+		if err := check.HealthCheck(ctx); err != nil {
+			errs = errors.Join(errs, err)
```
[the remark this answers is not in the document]
IfSentient: (or do the other suggestion)

health/observer.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+
+	return errs
+}
+
+func (c *Observer) CollectInitialChecks(ctx context.Context) error {
```
IfSentient, line 73: We can collect initial checks on the start of `Run`

operator/metrics_server.go
```diff
@@ -0,0 +1,111 @@
...
+			_, _ = w.Write([]byte("readiness check failed: " + err.Error()))
+			return
+		}
+		_, _ = w.Write([]byte("ok"))
+		return
+	}))
+	return nil
+}
+
+func (s *MetricsServer) Run(ctx context.Context) error {
```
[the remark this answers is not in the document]
charandas: I couldn't find a way of listening to Ctx.Done() on the parent context inside of a select without starting `ListenAndServe` inside a goroutine also.

The only way to use select, it seems, is to use a channel to read from. One could create an indirection from ListenAndServe into a channel, then wire a select similar to what you are stating above, but since select blocks, the running of the server itself needs to be in a goroutine. I will paste some scratch code below with waitgroup logic removed.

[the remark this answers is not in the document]
charandas:
```go
	observerCtx, cancelObserver := context.WithCancel(ctx)
	defer cancelObserver()

	go func() {
		s.observer.Run(observerCtx)
		_ = server.Shutdown(ctx)
	}()

	serverErr := make(chan error, 1)

	go func() {
		serverErr <- server.ListenAndServe()
	}()

	var returnedErr error
	select {
	case err := <-serverErr:
		returnedErr = err
	case <-ctx.Done():
		_ = server.Shutdown(ctx)
	}

	return returnedErr
```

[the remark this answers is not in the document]
charandas: So we might as well bring those waitgroups back in, I think.

radiohead (approved):
Nice work! I left a few comments but nothing blocking.

app/runner.go
```diff
@@ -348,6 +379,22 @@ func (d *DynamicMultiRunner) PrometheusCollectors() []prometheus.Collector {
 	return collectors
 }
 
+// HealthChecks implements health.Checker
+func (d *DynamicMultiRunner) HealthChecks() []health.Check {
```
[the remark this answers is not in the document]
IfSentient: Probably not a bad idea, not having it doesn't break functionality, but it could mean that we miss a runner which is added for a health check. 

Of course, if we only gather health checks at initialization (like prometheus exporters), then the `DynamicMultiRunner` will necessarily fall out of date (as a runner my be removed so its health check is no longer valid, or may be added later after health checks have been gathered). I think we'll want to work on a better solution for this specifically when handling health checks, but that should be left to the future for now.

IfSentient (approved):
lgtm

health/observer.go
```diff
@@ -0,0 +1,112 @@
...
+	// collect initial checks before we start listening to the ticker
+	initialStatus := c.collectChecks(ctx)
+	c.runLock.Lock()
+	c.runStatus = initialStatus
+	c.runLock.Unlock()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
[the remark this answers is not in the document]
charandas: Sorry, missed this. Will need to fix in a future PR.

PR #639, opened 2025-02-11
AppManifest Enhancements and Generated Code/Definitions

IfSentient:
Enhance AppManifest CUE definition with status information and dry-run information. Add make target (`make generate`) to generate go code for the appmanifest type, which can use used when interacting with the API server. Added non-generated conversion from the `v1alpha1.AppManifestSpec` type to `app.ManifestData`, for usage in `app.Provider` (and other places), allowing a consistent type to be used in `app` code, but an evolving API type to be used.

~Generated code currently has one hand-made change to name the `"*"` `AppManifestAdmissionOperation` `AppManifestAdmissionOperationAll`, related to https://github.com/grafana/cog/issues/757~ (we can use `@cog` in the CUE to manually name the enum values)

The non-generated files are `Makefile`, `app/manifest.cue`, and `app/appmanifest/v1alpha1/conversion.go`


app/manifest.cue
```diff
@@ -78,6 +87,25 @@ manifest: versions: v1alpha1: {
 				// accessKinds is a list of KindPermission objects for accessing additional kinds provided by other apps
 				accessKinds: [...#KindPermission]
 			}
+			// DryRunKinds dictates whether this revision should create/update CRD's from the provided kinds,
+			// Or simply validate and report errors in status.resources.crds.
+			// If dryRunKinds is true, CRD change validation will be skipped on ingress and reported in status instead.
+			// Even if no validation errors exist, CRDs will not be created or updated for a revision with dryRunKinds=true.
+			dryRunKinds?: bool | *false
```
IfSentient, line 94: I am debating whether this is better as-is, or if we want to add the ability to dry-run any resource types created/updated by the controller with the same structure as `status.resources`, like:
```cue
dryRun?: {
    [string]: bool
}
```
Or perhaps
```cue
dryRun?: {
    crds: bool | *false
    [string]: bool
}
```
Thoughts on this?

onprem: Not a strong opinion. I feel this is fine as is, more predictable and should be fine for most use cases I can think of.

onprem (approved):

app/appmanifest/v1alpha1/conversion.go
radiohead, line 1: Suggestion for the future – might be a good idea to add some tests based on golden values.

radiohead (approved):

app/manifest.cue
radiohead: IMHO kinds is fine for now.

radiohead (approved):

PR #658, opened 2025-02-19
[AppManifest] Add tests for generated AppManifest code and conversion code

IfSentient:
This is a follow-up to https://github.com/grafana/grafana-app-sdk/pull/639

Add tests for generated AppManifest code and conversion code. As part of this, I realized that some fields (`extraPermissions` and `selectableFields`) should be optional in the CUE, so as not to be required in the CRD, so I updated those fields and re-ran `make generate`.

### DIFF
--- a/app/manifest.cue
+++ b/app/manifest.cue
@@ -63,7 +63,7 @@ appManifest: versions: v1alpha1: {
 			name: string
 			admission?: #AdmissionCapabilities
 			schema: #ManifestKindVersionSchema
-			selectableFields: [...string]
+			selectableFields?: [...string]
 			additionalPrinterColumns?: [...#AdditionalPrinterColumns]
 		}
 		#ManifestKind: {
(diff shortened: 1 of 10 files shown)


app/manifest.cue
```diff
@@ -83,9 +83,9 @@ appManifest: versions: v1alpha1: {
 			kinds: [...#ManifestKind]
 			// ExtraPermissions contains additional permissions needed for an app's backend component to operate.
 			// Apps implicitly have all permissions for kinds they managed (defined in `kinds`).
-			extraPermissions: {
+			extraPermissions?: {
 				// accessKinds is a list of KindPermission objects for accessing additional kinds provided by other apps
-				accessKinds: [...#KindPermission]
+				accessKinds?: [...#KindPermission]
```
radiohead, line 88: I think this one shouldn't be optional, right? Otherwise we can have `extraPermissions` with empty content, which doesn't make much sense.

radiohead (approved):

app/manifest.cue
IfSentient: Good point, I'll take that off

radiohead (approved):

Issue #645, opened 2025-02-13
cue-based codegen generates incorrect app manifest and k8s CRD json for unstructured field


ellisda:
EXPECTED:
When we our `magefile codegen` as `sh.Run("grafana-app-sdk", "generate", "-f", "cue", "-t", "src/generated", "-c", "./schemas")` I expect the array type to generate manifest and k8s CRD that includes `"x-kubernetes-preserve-unknown-fields": true` and does not have an unspecified `"additionalProperties": {},` section

ACTUAL:
the grafanaQueries array gets generated as follows, and it has an extra `additionalProperties` and is missing the `x-kubernetes-preserve-unknown-fields: true`

```
                      "grafanaQueries": {
                        "items": {
                          "additionalProperties": {},
                          "type": "object"
                        },
                        "type": "array"
                      },
```

This is based on a cue file where we want to represent unknown key/value pairs directly instead of encoding our source json object as a string.

We're using a type 
```
	schema:
	{

		#GrafanaQueries: {
			grafanaQueries: [...#GrafanaQuery]
		} @cuetsy(kind="interface")

		#GrafanaQuery: {
			[string]: _
		} @cuetsy(kind="interface")
	    spec: {
	        foo:  #GrafanaQueries
	}

```

PR #659, opened 2025-02-20
[codegen] Fix bug in CRD generation where 'additionalProperties' is not parsed in arrays

IfSentient:
Resolves https://github.com/grafana/grafana-app-sdk/issues/645

Codegen for CRDs and schemas in the app manifest was not properly parsing openAPI array types to replace `"additionalProperties": {}` with `"x-kubernetes-preserve-unknown-fields": true`. 

The code that did this replacement (`replaceAdditionalProperties`) was a bit hard to follow, so I both fixed it there, and also introduced a new `setKubernetesPreserveUnknownFields` method which I think is a bit easier to follow and may be a bit more extensible. Please let me know if this is harder to follow, or if the previous, less-lines-of-code method feels like the better solution to continue with.

Added another test for this into the test CUE, and regenerated golden test files.


codegen/jennies/crd.go
```diff
@@ -210,17 +210,117 @@ func CUEToCRDOpenAPI(v cue.Value, name, version string) (map[string]any, error)
...
+	case "object":
+		setKubernetesPreserveUnknownFieldsInObject(castItems)
+		list["items"] = castItems
+	case "array":
+		setKubernetesPreserveUnknownFieldsInArray(castItems)
+		list["items"] = castItems
+	}
+}
+
+//nolint:unused
```
radiohead, line 308: I'm sorry, I can't seem to find where is this used from, could you point it out?

radiohead (approved):
NIT - IIUC there's an unused method that we can drop but not a blocking issue.

codegen/jennies/crd.go
IfSentient: It's not used, it's the old method that was used to replace the `additionalProperties`, which I fixed th ebug in, but found it to be overall a bit confusing so I replaced it with the new one. Left it in the PR in case folks thought that it was more straightforward than the new one that uses three functions instead.

IfSentient: I'll remove it now

radiohead (approved):

PR #679, opened 2025-03-06
Make operator status fields generation optional and remote status fields type prefixes

radiohead:
### What

This PR makes a few changes to how SDK generates status subresource:

1. Operator status fields generation can now be disabled via a flag
2. Status subresource is re-defined in the CUE overlay the same way as the spec (i.e. `_` or anything)
3. Subresource fields in Go generated object are no longer prefixed by their types (i.e. `Status DashboardStatus` instead of `DashboardStatus DashboardStatus`)

### Why

Most of these changes were made to satisfy the requirements of https://github.com/grafana/grafana/pull/99966 (code for which was generated using this branch).

More specifically:

1. Operator status fields are not always required.
2. Having status subresource defined as `[string]: _` results in cog not being able to follow type references and although the produced code still works, the generated types long auto-generated names (e.g. `Dashboarv1alpha1ConversionStatus` rather than just `DashboardConversionStatus`). It also makes status subresource an open struct (i.e. appends `additionalParameteres: {}` / `"x-kubernetes-preserve-unknown-fields": true`) _implicitly_.
3. For consistency with how `Spec` field is named (no type prefix) and consistency with vanilla Kubernetes types.


IfSentient (approved):
lgtm, I've been trying to work out a way to phase out the default status (or improve it in some way), I think this flag makes sense for doing that. I don't _think_ the `x-kubernetes-preserve-unknown-fields` being removed in the status object in the CRD should impact anything, I think it's an artifact of the source CUE having `[string]: _` joined to the default status.

`codegen/testing/golden_generated/go/groupbykind/customapp_manifest.go.txt` should be deleted for the integration testing to pass, or, at least, that used to be the case, but it looks like it's ok now. We might still want to remove it as it's unused and including it is a bug that needs to be fixed in the regenerate script.

toddtreece (approved):
the approach and changes make sense to me, ~~but it might be worth waiting for @IfSentient's feedback ~~

PR #700, opened 2025-03-13
Add a config option to allow empty disjunctions in Go

radiohead:
### What

This change adds a new config option, which allows for empty disjunctions in Go (using new cog codegen option) and wires it accordingly.

### Why

To allow SDK users to use this option - it is necessary when using disjunctions and integrating generated code directly into Kubernetes API server code (e.g. when building extension API servers).

Without this option generated code results in API server panics.


radiohead:
Depends on https://github.com/grafana/cog/pull/779 and new cog release

IfSentient (approved):

PR #710, opened 2025-03-17
Fix CopyObject to deep copy the object rather than shallow copy

IfSentient:
Fix CopyObject to deep copy the object rather than shallow copy, introduce CopyObjectInto to do a full deep copy using reflection. Add tests for CopyObject and CopyObjectInto. Add DeepCopy and DeepCopyInto methods to generated object and spec/subresource code.

### DIFF
--- a/resource/object.go
+++ b/resource/object.go
@@ -179,15 +179,19 @@ func CopyObjectInto[T any](out T, in T) error {
 	dstVal := reflect.ValueOf(out)
 	srcType := reflect.TypeOf(in)
 	dstType := reflect.TypeOf(out)
+	// T must be a pointer to a struct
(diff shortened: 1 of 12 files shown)


resource/object.go
```diff
@@ -149,16 +150,141 @@ type CommonMetadata struct {
 //	    return resource.CopyObject(c)
 //	}
 func CopyObject(in any) Object {
```
ryantxu, line 152: Should this be renamed to `DeepCopyObject` so we are clear?  (and keep a CopyObject that is not deep?)

IfSentient: I'm not sure we need/want a CopyObject that isn't deep, or rather, if we do, we might want it to be clearer from the naming that it's a shallow copy.

PR #702, opened 2025-03-13
adjust how custom routes are handled

toddtreece:

PR #723, opened 2025-03-26
Add CustomRoutes to App Manifest Go Codegen

toddtreece:
* `app/manifest.go`: added `CustomRoutes` to `ManifestKindVersion`
* `codegen/cuekind/def.cue`: added new definitions for `PathProps`, `Operation`, and related openapi v3 components to support custom route definitions


codegen/cuekind/testing/testkind.cue
```diff
@@ -48,5 +48,176 @@ testKind: {
...
+				spec: {
+					stringField: string
+					intField: int64
+					timeField: string & time.Time
+					boolField: bool
+				}
+			}
+			mutation: operations: ["create","update"]
+			validation: operations: ["create","update"]
+			customRoutes: {
```
radiohead, line 62: My $0.02 – I think this is great but I find it way too verbose and low-level.

I guess that's what k8s expects but I would ideally like us to abstract & simplify (even if it means losing some fidelity / flexibility) this.

As a developer I personally wouldn't want to be this verbose, plus it's very easy to accidentally write some invalid configuration.

If we are not under time pressure, I would love to brainstorm a better, more high-level approach for expressing this and maybe put together a design doc for this feature.

Also, ideally whatever configuration I put into custom routes, we should be generating handlers for in Go, with proper parsing of requests & forming of responses, including things like request type structs, etc, etc, based on e.g. https://github.com/oapi-codegen/oapi-codegen.

@IfSentient WDYT?

toddtreece: > As a developer I personally wouldn't want to be this verbose, plus it's very easy to accidentally write some invalid configuration.

i think the cue types currently verify the structure, but could you clarify what you mean by invalid here? happy to adjust if there are suggestions on how to make it less verbose. unlike kind schemas there are additional properties related to the request/response, so following a known format seemed like a reasonable path forward in terms of supporting everything that a user might want to document about their custom routes. to be honest i spent a lot of time with this getting used to more complex codegen changes in this repo, so i have not considered alternate schema approaches.

> Also, ideally whatever configuration I put into custom routes, we should be generating handlers for in Go

i had something like that in #702 but i figured that could be handled separately because it's a different part of the workflow

radiohead: > i think the cue types currently verify the structure, but could you clarify what you mean by invalid here?

IIUC I could define the same path twice with the same config and that'd still be valid, right? e.g. `{ "/reconcile": { get: {}}, "/reconcile": { get: {}} }`. I'm sure there's more footguns here 😅 

> a reasonable path forward in terms of supporting everything that a user might want to document about their custom routes

Yes, I agree.

I'm just coming from a different design philosophy of supporting fewer use-cases at the cost of less complexity, as long as that's OK. If possible I'd rather start by adding the bare minimum and grow the feature set as we go, once we have the necessary use-cases.

My rationale here is that a good UX with no (or very few) footguns leads to fewer bugs and constraints in the amount of features can lead to better designs. Think Go vs. Scala design.

> i had something like that in https://github.com/grafana/grafana-app-sdk/pull/702 but i figured that could be handled separately because it's a different part of the workflow

Okay, fair enough. I'm OK with handling the codegen / SDK changes separately.

toddtreece: > that'd still be valid, right? e.g. `{ "/reconcile": { get: {}}, "/reconcile": { get: {}} }`

i'd assume for this example that the non-unique path keys would be an issue in cue

> I'm just coming from a different design philosophy of supporting fewer use-cases at the cost of less complexity, as long as that's OK

@radiohead sounds reasonable to me. one thing to note is that these types of routes already exist in data sources and app plugins, so in some cases it's more about documenting routes that might already have existing complexity (especially if they are mostly used as a reverse proxy to another API). it might also be worth thinking about if the rough shape of a fully featured end state would end up in the same spot. i'll follow up in slack

toddtreece:
closing in favor of #727

PR #724, opened 2025-04-01
operator: Add initial concurrent reconciliation support

onprem:
### What

This PR adds support for reconciling events concurrently in `operator.InformerController`. By default only `1` worker would be run, to maintain backward compatibility of processing the events sequentially.

This is done with the help of a new `ResourceWatcher` implementation called  `ConcurrentWatcher`, which runs multiple workers in the background and shards the incoming events between them. For every event, a hash mod is calculated on the identifier, allowing us to deliver events for same object to the same worker. This is done to maintain the per object in-order guarantee for events by Kubernetes.

### Why

This is needed for https://github.com/grafana/cloud-apps-platform/issues/883

### DIFF
--- a/operator/concurrentwatcher.go
+++ b/operator/concurrentwatcher.go
@@ -2,20 +2,26 @@ package operator
 
 import (
 	"context"
+	"fmt"
 	"sync"
+	"time"
 
 	"github.com/cespare/xxhash/v2"
 	"github.com/grafana/grafana-app-sdk/resource"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/buffer"
 )
 
 var (
-	defaultBuffer = 1000
+	initialBufferSize = 1024
 )
 
 type eventInfo struct {
+	ctx    context.Context
 	action ResourceAction
 	target resource.Object
-	old    resource.Object
+	source resource.Object
 }
 
 type worker struct {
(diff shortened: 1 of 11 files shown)


operator/informer_controller.go
```diff
@@ -239,20 +254,128 @@ func (c *InformerController) AddInformer(informer Informer, resourceKind string)
 		return fmt.Errorf("resourceKind cannot be empty")
 	}
 
-	err := informer.AddEventHandler(&SimpleWatcher{
-		AddFunc:    c.informerAddFunc(resourceKind),
-		UpdateFunc: c.informerUpdateFunc(resourceKind),
-		DeleteFunc: c.informerDeleteFunc(resourceKind),
-	})
+	watcher := NewConcurrentWatcher(
```
IfSentient, line 257: We may want to only use the ConcurrentWatcher if `c.informerMaxWorkers > 1`, so we don't go through additional logic if the app doesn't need/want concurrency.

IfSentient (commented):
Looks good! Could we add some tests for `ConcurrentWatcher` since it's in the path for handling all events going forward?

operator/informer_controller.go
onprem: Good point. It would also help us avoid some memory overhead of an additional buffer too. Plus cleaner spans, simpler logic.

Will do that and also add tests for the `ConcurrentWatcher`.

onprem:
> Looks good! Could we add some tests for `ConcurrentWatcher` since it's in the path for handling all events going forward?

I have added tests for `ConcurrentWatcher`, and cleaned up a few more things.

operator/concurrentwatcher.go
```diff
@@ -0,0 +1,157 @@
...
+	source resource.Object
+}
+
+// ConcurrentWatcher is a struct that implements ResourceWatcher, but takes no action on its own.
+// For each method in (Add, Update, Delete) the event is added in a buffered queue and the corresponding
+// methods of the underlying ResourceWatcher are called concurrently.
+// The events are processed by a fixed number of workers running concurrently. Each worker processing
+// the events one-by-one. The events are sharded using hash mod algorithm, ensuring events for same object
+// end up in the same worker. This is to uphold the guarantee of in-order delivery of events for an object.
+type ConcurrentWatcher struct {
```
IfSentient, line 33: A thought when looking this over again--this might be better as a wrapper for an informer, rather than a watcher, as it requires being `Run` like an informer (watchers tend to be passive), and we use it to wrap an informer's event handler function anyway when adding on the informer side of the `InformerController`. If we revert the `Informer` interface to still take a `resource.Object`, we can still hashMod the same way as well. Since logically, this is added on the informer-add end of things, rather than the watcher-add (like with wrapping a watcher with `OpinionatedWatcher`), I'm curious if naming this as an informer (and adhering to the `Informer` interface) would be a better logical fit for it. Thoughts?

IfSentient (commented):
A few new comments after going through everything again, sorry for the delay.

operator/concurrentwatcher.go
onprem: I totally agree. It fits more as an Informer wrapper, especially with the need to manage the life-cycle for underlying workers.

onprem: I have updated the code to not leak the concurrency related logic into the `InformerController`, and instead the choice of abstraction is now a new `ConcurrentInformer` that wraps and existing informer.

onprem:
I have moved the abstraction out of `InformerController` to a separate Informer implementation as discussed. This resulted in cleaner code, and better isolation of logic.

operator/informer_concurrent.go
```diff
@@ -0,0 +1,88 @@
...
+	informer             Informer
+	watchers             []*concurrentWatcher
+	maxConcurrentWorkers uint64
+
+	mtx sync.RWMutex
+}
+
+type ConcurrentInformerOptions struct {
+	// ErrorHandler is a user-specified error handling function. This is typically for logging/metrics use,
+	// as retry logic is covered by the RetryPolicy. If left nil, DefaultErrorHandler will be used.
```
IfSentient, line 27: NIT: `RetryPolicy` only exists for the `InformerController`, so we may want to omit mention of it here.

IfSentient (approved):
This looks great! I think starting with this behavior needing to be opted-into by specifying the informer is good, and once we've confirmed this works well in production we can perhaps have it be added by default in apps similar to `OpinionatedWatcher`/`OpinionatedReconciler` wrapping watchers/reconcilers by default in `simple.App`.

ferruvich (approved):

PR #727, opened 2025-04-01
App Manifest Custom Sub-Resource Routes

toddtreece:
## What
Adds support for defining custom sub-resource routes in app manifests. This is the foundation for supporting custom HTTP endpoints in apps. Handling will be implemented in a future PR.

## Why
Some apps need to expose dataplane related routes outside of the existing k8s resource routes (CRUD + list + watch).


IfSentient (approved):
This looks great! I think in future passes we'll want to push more validation into the CUE/parser, but as a first pass to get things working and a sensible CUE definition for codegen'ing custom subresource routes, this is good.

PR #729, opened 2025-04-02
[MemcachedStore] Allow for specifying a ServerSelector and automatic server refresh

IfSentient:
Allow for specifying a `ServerSelector` in the `MemcachedStore` and automatically refresh the servers when an EOF or timeout is encountered.

A default implementation of the `MemcachedServerSelector` using `memcache.ServerList` has been added to the `simple` package.


simple/memcached.go
```diff
@@ -0,0 +1,48 @@
+package simple
```
IfSentient, line 1: I am on the fence about this being in the `simple` package, but it didn't feel like it belonged in the `operator` package either.

IfSentient:
CC @ferruvich

operator/cache_memcached.go
```diff
@@ -253,7 +281,11 @@ func (m *MemcachedStore) GetByKey(key string) (item any, exists bool, err error)
 
 func (m *MemcachedStore) getByKey(key string) (item any, exists bool, err error) {
 	start := time.Now()
-	fromCache, err := m.client.Get(key)
+	var fromCache *memcache.Item
+	err = m.attemptWithRefreshOnTimeout(func() error {
```
ferruvich, line 285: one question: should the `attemptWithRefreshOnTimeout` be used **each** time the `m.client` is used? Because I can see some "plain" calls to `m.client` in this file 🤔

IfSentient: Ah, you're right, I missed doing this with the key sync stuff for handling cache resyncs. Let me push a fix for that.

ferruvich:
@IfSentient is this ready for review again?

IfSentient:
> @IfSentient is this ready for review again?

@ferruvich Ah, yup, sorry that I missed this ping

ferruvich (approved):

PR #730, opened 2025-04-03
Handle k8s API status error types correctly

radiohead:
### What

This change modifies k8s error parsing in clients to properly ingest API status errors, rather than trying to cobble together information from the HTTP status and messages.

### Why

This makes detailed error information from API accessible by clients (e.g. field validation errors), which is required for good end-user UX.


radiohead:
Example of field validation:

* Populated by server in https://github.com/grafana/grafana/pull/103293/files#diff-a93108dbfd13fd5a22dba1af8e863c28dc92180beced9b8d0586ad1fd1fe9a40R90
* Read by the client in https://github.com/grafana/terraform-provider-grafana/pull/2113/files#diff-54047a5d286e4407b283eb7f76b433701c0222362826c8084d6fd4704f1c18f1R26

k8s/gvclient.go
```diff
@@ -617,29 +617,28 @@ type k8sErrBody struct {
...
-	}
 	if len(responseBytes) > 0 {
-		parsed := k8sErrBody{}
+		var parsed metav1.Status
 		// If we can parse the response body, use the error contained there instead, because it's clearer
 		if e := json.Unmarshal(responseBytes, &parsed); e == nil {
-			err = errors.New(parsed.Message)
-			if statusCode == 0 && parsed.Code > 0 {
-				statusCode = parsed.Code
+			return &k8serrors.StatusError{
```
IfSentient, line 624: We use the `StatusCode()` method on the various resource Stores to check if a response is a 404 (for upsert, [example](https://github.com/grafana/grafana-app-sdk/blob/main/resource/typedstore.go#L125)), so we'll need to make changes there as well, or wrap both `ServerResponseError` and `StatusError` with an interface that presents the same information.

radiohead: OK, I can take care of that.

radiohead: Updated. I've ended up refactoring error handling in a pretty major way, but on the upside all error handling should now be compatible with k8s api errors package.

IfSentient (approved):

k8s/gvclient.go
IfSentient: Looks good, I think the error refactor is cleaner for handling everything!

PR #739, opened 2025-04-11
[AppManifest] Add operator information to the AppManifest

IfSentient:
Add operator information to the AppManifest, to allow for automatic configuration of webhooks for operators.

The information in the manifest data itself is more extensible than what is processed in codegen (manifest data has a base URL, then paths for each webhook capability, while codegen just accepts the operator URL and auto-generates the paths based on the paths automatically used by `operator.Runner`/[k8s.WebhookServer](https://github.com/grafana/grafana-app-sdk/blob/main/k8s/webhooks.go#L153-L155)), as without a custom operator runner, the webhook paths are immutable. This may change in the future, and the codegen input manifest data can be extended.


ferruvich (approved):
LGTM 👍🏼 
I think it's ok to enforce the validation/mutation/conversion path automatically used by `operator.Runner` - I like having a standard paths for serving these operations.

PR #775, opened 2025-04-30
[AppManifest] Add generated function to associate a kind and version to its go `resource.Kind`

IfSentient:
To build an apiserver runner, the go types for an app must be registered in the scheme prior to the point of the loopback interface creation to provide as a kubeConfig to the `NewApp` function. This creates a slight chicken-and-egg problem, where the app is the only thing that knows the specific go types for the kinds it manages, but needs a kubeConfig to be initialized for `ManagedKinds()` to be called by the runner. The AppManifest data (`app.ManifestData`) has all the information on the kinds managed by the app, _excepting_ the go types it uses, as the manifest data is assumed to be file-like, and can be loaded from disk or an apiserver as well.

This PR is a proposal for a solution to this problem, by adding a generated function alongside the generated manifest that associates a kind (string) and version (string) to the app's corresponding generated `resource.Kind` implementation. The apiserver runner config can than consume that function in its `New` method, and allow it to make the associations based on the manifest for the app prior to loading the app itself. If an app author does not use the codegen, they can simply implement this function as well if using the apiserver runner.

~This PR is a proposal for a solution to this problem, by extending the AppManifest as it is represented in-code.~

~To do this, this PR adds an association in the `app.ManifestData` to the go `resource.Kind` for each KindVersion, and adds extra code and config to the only extant runner, `operator.Runner`, to be able to make the `app.ManifestKindVersion`-to-`resource.Kind` association after loading the manifest data from a file (if an embedded, in-go manifest is used, it does not try to do this). Codegen has been updated to add the association in the embedded manifest, and add a new exported function that handles doing the `Kind`+`Version` => `resource.Kind` association for a runner which loads a manifest from disk or elsewhere to hook into.~


IfSentient:
CC @toddtreece

app/manifest.go
```diff
@@ -109,6 +111,8 @@ type ManifestKindVersion struct {
 	SelectableFields []string `json:"selectableFields,omitempty" yaml:"selectableFields,omitempty"`
 	// CustomRoutes is a map of of path patterns to custom routes for this version.
 	CustomRoutes map[string]spec3.PathProps `json:"customRoutes,omitempty" yaml:"customRoutes,omitempty"`
+	// GoKind is the association of this kind to the go resource.Kind type, which can be used for accessing the go type(s) and codec information
+	GoKind resource.Kind `json:"-" yaml:"-"`
```
toddtreece, line 115: would it be easier/better to use the `ManifestGoTypeAssociator` util function for now and avoid modifying `ManifestKindVersion`?

IfSentient: I think this might be the best approach, no need to make it more complicated right now with the addition of the AppManifest changes. The apiserver runner would just require the association function in its config so it can extract go types from the manifest.

toddtreece: sounds good to me

IfSentient: Alright, updated the PR to remove the `GoType` association in the manifest (and updated the description).

PR #779, opened 2025-05-02
[AppManifest] Implement `app.VersionSchema.AsKubeOpenAPI`

IfSentient:
To move toward enabling apiserver functionality without needing the `--postprocess` flag when generating code, we need to be able to provide a function that consumes a kubernetes `kube-openapi` `common.ReferenceCallback` and returns a `map[string]common.OpenAPIDefinition`. Currently, this is generated by using kubernetes `openapi-gen` in the jenny that runs for `--postprocess`, but this sources from generated go code, and information can be lost in the conversion from CUE->go->OpenAPI. Additionally, we would like to be able to source everything required for an app to run from the AppManifest itself, and not require extra generated code for the runner.

This PR implements the function `AsKubeOpenAPI` on the AppManifest's `VersionSchema` object, which produces similar output to the output created by generated code. This allows an apiserver runner to require only the manifest for getting the required kubernetes openAPI objects, rather than needing to reference generated code.


app/manifest.go
```diff
@@ -173,6 +179,17 @@ type KindPermission struct {
 	Actions  []KindPermissionAction `json:"actions,omitempty" yaml:"actions,omitempty"`
 }
 
+type ManifestOperatorInfo struct {
```
ferruvich, line 182: nit: let's add some godoc on this and the struct below as they're exported

app/manifest.go
```diff
@@ -289,18 +306,281 @@ func (v *VersionSchema) AsOpenAPI3() (*openapi3.Components, error) {
...
+		},
+		Dependencies: make([]string, 0),
+	}
+
+	// For each schema, create an entry in the result
+	for k, s := range oapi.Schemas {
+		sch, deps := oapi3SchemaToKubeSchema(s, ref, gvk)
+		// sort dependencies for consistent output
+		slices.Sort(deps)
+		result[fmt.Sprintf("%s/%s.%s", gvk.Group, gvk.Version, k)] = common.OpenAPIDefinition{
```
ferruvich, line 370: very nit: consider declaring `fmt.Sprintf("%s/%s.%s", gvk.Group, gvk.Version, k)` separately as I can see it's repeated some times in this for loop

app/manifest.go
IfSentient: Ah, yeah, I should have done that in a prior PR, glad the moving of this section brought this up. Will do.

IfSentient:
CC @toddtreece on this as well, as this contributes to apiserver support in the app-sdk.

toddtreece (approved):
nice, thanks!

PR #784, opened 2025-05-06
[Proposal] Updates to `app.App.CallResourceCustomRoute` to handle other custom route types and response streaming

IfSentient:
This PR updates `app.App.CallResourceCustomRoute` to `app.App.CallCustomRoute`, and enhances the request and response to handle both:
* resource routes (routes that aren't a subresource of a kind) and subresource routes
* response streaming (with docs on how a runner should implement the streaming or non-streaming)

`simple.App` has been updated to handle both of these changes, though registration of resource routes (non-subresource routes) is not yet supported in `simple.App`.


app/app.go
```diff
@@ -35,19 +36,40 @@ type RawObject struct {
...
 	StatusCode int
-	Body       []byte
+	// Body is the payload of the response. Runners MAY support response streaming.
+	// In this case, an EOF returned by Body.Read() should designate the end of the response
+	// (0 bytes read with no error only denotes that there is nothing currently to write to the response stream),
+	// consistent with the behavior described in the godoc for io.Reader.
+	// If a Runner does not support response streaming, it should call Close() on Body when the payload's response
+	// has been written and the stream closed (this should also be done when streaming is supported after getting an
+	// EOF or other error and no further bytes).
+	Body io.ReadCloser
```
IfSentient, line 72: The other solution here (as pointed out by @toddtreece) would be to provide a writer/sender object in the request object or `CallCustomRoute` method, and the app can keep the response stream open by not returning from the method. I think that would move us more toward following the `http.ResponseWriter` approach, which we had initially and changed to a concrete response object. I'd love opinions on which of these two approaches we think may be better.

simple/app.go
```diff
@@ -500,14 +503,27 @@ func (a *App) Convert(_ context.Context, req app.ConversionRequest) (*app.RawObj
...
-func (a *App) CallResourceCustomRoute(ctx context.Context, req *app.ResourceCustomRouteRequest) (*app.ResourceCustomRouteResponse, error) {
-	k, ok := a.kinds[gvk(req.ResourceIdentifier.Group, req.ResourceIdentifier.Version, req.ResourceIdentifier.Kind)]
+// CallCustomRoute implements app.App and handles custom resource route requests
+func (a *App) CallCustomRoute(ctx context.Context, req *app.CustomRouteRequest) (*app.CustomRouteResponse, error) {
+	if req.ResourceIdentifier.Kind == "" && req.ResourceIdentifier.Plural == "" {
+		scope := resource.NamespacedScope
+		if req.ResourceIdentifier.Namespace == "" {
+			scope = resource.ClusterScope
+		}
+		if handler, ok := a.customRoutes[a.customRouteHandlerKey(nil, req.Method, req.Path, scope)]; ok {
```
toddtreece, line 513: do resource level custom routes also need to be added to `AppConfig`?

IfSentient: Yeah, this PR doesn't aim to support them yet in `simple.App`, but we'll need a way to add them, which would have to be at the `AppConfig` level.

app/app.go
toddtreece: the writer/sender approach seems slightly easier for users that only need to send `[]byte`, but wrapping with `io.NopCloser(bytes.NewReader([]byte{}))` isn't that huge of an issue, so either way works for me.

this seems like an edge case, but the only thing i can think of is that sending trailer headers wouldn't be supported with `io.ReadCloser` (although maybe you could write them to the body? not sure)

app/app.go
```diff
@@ -35,19 +35,46 @@ type RawObject struct {
...
+	// Headers contains the HTTP headers for the response
+	Headers http.Header
+	// Body is the contents of the response
+	Body []byte
+}
+
+var ErrCustomRouteResponseWriterClosed = errors.New("custom route response writer is closed")
+
+// CustomRouteResponseWriter is a ResponseWriter for CustomRouteResponse objects.
+type CustomRouteResponseWriter interface {
```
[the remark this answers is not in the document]
IfSentient: @toddtreece @radiohead thoughts on the interface here?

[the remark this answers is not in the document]
radiohead: IMHO
```golang
type CustomRouteResponseWriter interface {
    http.ResponseWriter
}
```
makes the most sense for now.

PR #795, opened 2025-05-13
[AppManifest][Codegen] Use Version -> Kind rather than Kind -> Version

IfSentient:
**Note:** This is a rather large PR which may need to be split into several, but this contains everything necessary for the new functionality (AppManifest changes, codegen/parser changes, CLI changes).

# What This PR Does / Why We Need It

In kubernetes, kinds exist in a hierarchy that can be seen in the URL to access one, with `apigroup` being the top of the hierarchy, `apiversion` being the next level, and finally the kind itself being the lowest level. As such, while kinds belong to an `apigroup` (which is analogous to an App within App Platform contexts), they _also_ are localized to an `apiversion`, despite a kind with the same name (and other metadata, like scope) possibly existing in another version. This is a bit confusing, because kind data and `apiversion` are somewhat entangled: a kind shares the same information between API versions, excepting its schema (and data tied to the schema, such as table view fields and field selectors). 

Representing this kind hierarchy in a way that is understandable and usable in the manifest becomes somehwhat of a challenge. The current model is to have the manifest contain a list of kinds (`manifest.kinds`), and the kinds themselves have a list of versions they exist for. However, since APIs are versioned and the preferred use version for API discovery is at the API group + version level, and not per-kind, an App and kind's versioning should be tied together as a single vertical. This vertical looks something like:

| Version | Kinds available in version |
|-|-|
| foo.ext.grafana.app/v1 | foos, bars |
| foo.ext.grafana.app/v2 | foos, bars, foobars |
| foo.ext.grafana.app/v3 | foos, bars, foobars, barfoos |

Note that new kinds introduced do not start at `v1`, despite the current model of kind-centric versioning. Instead, new kinds _may_ exist from the earliest (or earlier) versions, but must always exist in subsequent versions, regardless of whether that particular kind has schema changes that would warrant a new version. This is to present a uniform API surface per version, consistent with behavior of kubernetes APIs and service discovery.

This PR updates the way the App Manifest is defined (both in-code/as an API object and in codegen) to put versions at a higher level than kinds--that is, it inverts the relationship of manifest->kind(s)->version(s) to be manifest->version(s)->kind(s). Each kind in the manifest now if version-specific, rather than version-agnostic with version-specific attributes (like schema) found in `versions`. A manifest that may have previously been represented as
```cue
{
    appName: "test"
    kinds: [{
        kind: "Foo"
        current: "v2"
        versions: {
            "v1": {
                schema: { ... }
            }
            "v2": {
                schema: { ... }
            }
        }
    }]
}
```
Would now look like:
```cue
{
   appName: "test"
   versions: {
       "v1": {
           kinds: [{
               kind: "Foo"
               schema: { ... }
           }]
       }
       "v2": {
           kinds: [{
               kind: "Foo"
               schema: { ... }
           }]
       }
   }
   preferredVersion: "v2" // this can now be omitted and defaults to latest
}
```

This does introduce redundancy that didn't exist before in the kind-centric model. To get around this, the complexity of the CUE used for codegen can be modified (and the new generated CUE does this)
```cue
{
   appName: "test"
   versions: {
       "v1": {
           kinds: [foov1]
       }
       "v2": {
           kinds: [foov2]
       }
   }
}

fooKind: {
    kind: "Foo"
}

foov1: fooKind & {
    schema: { ... }
}

foov2: fooKind & {
    schema: { ... }
}
```

Due to the severity of this change, a flag has been added to the CLI for the `generate` and `project local generate` commands to parse the "old" style manifest (`--useoldmanifestkinds`), while still _producing_ the updated type of manifest for the output in go/json/yaml.

This change _does not_ impact the way `resource.Object` and `resource.Kind`s behave, as within the app-sdk's libraries, a kind is always treated as an entity with a unique group, version, and kind (which more closely mirrors the entry found in `manifest.versions[].kinds[]` in this new paradigm).

### Changes in `project` commands

A newly-initialized project now has the updated version of the manifest and a default `v1alpha1` version. Running `project kind add` now adds a kind to the _latest_ version in the version map, and uses the kind CUE strategy outlined above, with a file `<kindname>.cue` containing the kind metadata like so:
```cue
package kinds

// This is our Foo definition, which contains kind metadata. It is the same across all versions of the kind.
fooKind: {
    // Name is the human-readable name which is used for generated type names.
    kind: "Foo"
    // Scope determines the scope of the kind in the API server. It currently allows two values:
    // * Namespaced - resources for this kind are created inside namespaces
    // * Cluster - resource for this kind are always cluster-wide (this can be thought of as a "global" namespace)
    // If not present, this defaults to "Namespaced"
    scope: "Namespaced"
    // [OPTIONAL]
    // The human-readable plural form of the "name" field.
    // Will default to <name>+"s" if not present.
    pluralName: "Foos"
    // [OPTIONAL]
    // Codegen is a trait that tells the grafana-app-sdk, or other code generation tooling, how to process this kind.
    // If not present, default values within the codegen trait are used.
    // If you wish to specify codegen per-version, put this section in the version's object
    // (for example, foov1alpha1) instead.
    codegen: {
        // [OPTIONAL]
        // ts contains TypeScript code generation properties for the kind
        ts: {
            // [OPTIONAL]
            // enabled indicates whether the CLI should generate front-end TypeScript code for the kind.
            // Defaults to true if not present.
            enabled: true
        }
        // [OPTIONAL]
        // go contains go code generation properties for the kind
        go: {
            // [OPTIONAL]
            // enabled indicates whether the CLI should generate back-end go code for the kind.
            // Defaults to true if not present.
            enabled: true
        }
    }
}
```
and `<kind>_<version>.cue` containing the version-specific kind like so:
```cue
package kinds

// This is the v1alpha1 version of the kind, which joins the kind metadata and
// version-specific information for the kind, such as the schema
foov1alpha1: fooKind & {
    // schema is the schema for this version of the kind
    // ... truncated comment
    schema: {
        // spec is the schema of our resource. The spec should include all the user-editable information for the kind.
        spec: {
            // Example fields
            firstField: string
            secondField: int
        }
        // ... truncated comment
    }
}
```
Adding the kind to the manifest is a bit more complex than previously, and uses some brittle regex (just as its predecessor did), because the CUE can't be modified by the CUE sdk without altering the way the user wrote it. This has some room for improvement, but is no worse than the current functionality.

An additional `project version add` command has been introduced that adds a new version to the manifest and adds kinds to it based on the kinds for the current version.


cmd/grafana-app-sdk/project.go
```diff
@@ -78,13 +87,18 @@ func setupProjectCmd() {
 	projectAddComponentCmd.Flags().String("grouping", kindGroupingKind, `Kind go package grouping.
 Allowed values are 'group' and 'kind'. This should match the flag used in the 'generate' command`)
 
+	projectLocalGenerateCmd.Flags().Bool("useoldmanifestkinds", false, "Whether to use the legacy manifest style of 'kinds' in the manifest, and 'versions' in each kind. This is a deprecated feature that will be removed in a future release.")
+	projectLocalGenerateCmd.Flags().Lookup("useoldmanifestkinds").NoOptDefVal = "true"
```
toddtreece, line 91: is this going to be a lot to maintain? i'm wondering if it's easier to update the apps that use it in grafana one PR instead of adding a new flag and delaying the update

toddtreece (approved):
looks good to me, thanks!

radiohead (approved):

IfSentient:
Re-targeting to a feature branch that requires PRs for changes so custom routes and apiserver work can be done together and then merged once the whole feature is complete, to avoid excess impact.

PR #799, opened 2025-05-16
Chore: Update tutorial docs

Clarity-89:
Fix minor spelling issues and typos after following the tutorial.


docs/tutorials/issue-tracker/05-local-deployment.md
```diff
@@ -172,16 +172,15 @@ If we click on the `grafana` tile to see what the error is, it shows us the logs
 logger=provisioning t=2025-01-30T16:53:00.36954668Z level=error msg="Failed to provision plugins" error="app provisioning error: plugin not installed: \"issuetrackerproject-app\""
 ```
 
-So this is just something we didn't do before we started our environment. While we built the plugin, we want to also deploy it to our local cluster. The local cluster can only read from `local/mounted-files`, so we need it there. Simple enough fix, we run
+So this is just something we didn't do before starting our environment. When we build the plugin, we also need to deploy it to the local cluster. Because the cluster can only read from `local/mounted-files`, the plugin must be copied there. The fix is simple: run `make local/deploy_plugin` (in a separate terminal window).
+
```
Clarity-89, line 176: I got confused here as I first stopped the running `make local/up` and tried to run `make local/deploy_plugin` instead, which failed because the `make local/up` was not running.

docs/tutorials/issue-tracker/06-frontend.md
```diff
@@ -66,35 +66,57 @@ issue.spec.status = newStatus;
 ```
 
 Finally, we have to display the information. We define one more function:
-```TypeScript
+```tsx
 const getActions = (issue: Issue) => {
-    if(issue.spec.status === 'open') {
+    if (issue.spec.status === 'open') {
```
IfSentient, line 71: This whole section is actually just copied verbatim from [the file you add itself](https://github.com/grafana/grafana-app-sdk/blob/main/docs/tutorials/issue-tracker/frontend-files/main.tsx#L71), though I get the intent of making it easier to read in the doc.

IfSentient: 👍 , I'll make sure the TS in the source files gets formatted in the same consistent way.

PR #831, opened 2025-06-13
docs: refresh docs and minor changes

konsalex:
After running the tutorial found a couple of points to refresh the docs.

1. Updated the minio version to the latest, as the logs on the pod were stating:

```
You are running an older version of MinIO released 3 years ago. Update: Run `mc admin update`.
```

2. Minor typos in code examples in markdown files
3. Updated internal markdown links that were breaking
4. Changed the return of the example CRD conversion function, to follow the [K8s example](https://github.com/kubernetes/kubernetes/blob/v1.25.3/test/images/agnhost/crd-conversion-webhook/converter/example_converter.go#L36)
5. Remove Thema reference that was obsolete in the kubernetes.md

### DIFF
--- a/docs/admission-control.md
+++ b/docs/admission-control.md
@@ -21,7 +21,7 @@ func main() {
     validatingController := resource.SimpleValidatingAdmissionController{
         ValidateFunc: func(ctx context.Context, request *resource.AdmissionRequest) error {
             // Check that the name is allowed
-            if request.Object.StticMetadata().Name == "not-allowed" {
+            if request.Object.GetStaticMetadata().Name == "not-allowed" {
                 return k8s.NewSimpleAdmissionError(fmt.Errorf("name not allowed"), http.StatusBadRequest, "ERR_NAME_NOT_ALLOWED")
             }
             return nil
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -26,13 +26,11 @@ Custom Resource Definitions must have a **group** and **version**, which are use
 ### CRDs & grafana-app-sdk
 
 When you author kinds and use `grafana-app-sdk generate` to [generate code](code-generation.md), you'll get a CRD file for each kind as well. 
-The CRD file will contain the entire lineage expressed in openAPI format, and can be used to create your kind as a CRD in a kubernetes cluster. 
+The CRD file will contain the entire lineage expressed in openAPI format, and can be used to create your kind as a CRD in a kubernetes cluster.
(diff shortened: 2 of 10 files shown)


docs/admission-control.md
```diff
@@ -21,7 +21,7 @@ func main() {
     validatingController := resource.SimpleValidatingAdmissionController{
         ValidateFunc: func(ctx context.Context, request *resource.AdmissionRequest) error {
             // Check that the name is allowed
-            if request.Object.StticMetadata().Name == "not-allowed" {
+            if request.Object.GetStaticMetadata().Name == "not-allowed" {
```
IfSentient, line 24: Might be useful just to change this completely away from using `StaticMetadata`, as we now have getters for all metadata properties:
```suggestion
            if request.Object.GetName() == "not-allowed" {
```

docs/kubernetes.md
```diff
@@ -26,13 +26,11 @@ Custom Resource Definitions must have a **group** and **version**, which are use
...
+The CRD file will contain the entire lineage expressed in openAPI format, and can be used to create your kind as a CRD in a kubernetes cluster.
 The generated `resource.Schema` is then used to identify the **group**, **version**, and **kind** of your CRD when interacting with kubernetes. 
 You can then use a `k8s.ClientRegistry` to generate clients which can translate the CRDs in the cluster into the generated go type. 
 
-The reason the client intermediary is necessary is twofold:
-1. It allows thema to be inserted at the unmarshal point
-2. The format of metadata in your authored kind doesn't exactly match kubernetes' metadata format--the kind has more metadata, which is encoded and decoded in kubernetes' annotations for the object. The client performs this translation.
+The client intermediary is necessary because the metadata format in your authored kind doesn't exactly match Kubernetes' metadata format. The kind has more metadata, which is encoded and decoded in the Kubernetes object's annotations, and the client performs this translation.
```
IfSentient, line 33: This is somewhat out-of-date, as the client no longer encodes/decodes the annotations metadata, the generated object implementation has getters/setters for the metadata fields that encode it to the annotations. The generated object now should match the kubernetes expectation, but we use the client for two main reasons:
1. It allows us to introduce efficiencies under the hood transparently to the app author
2. It ensures that we use the encoding/decoding process in the kind's `resource.Codec`, rather than a straight JSON marshal/unmarshal

At this point, the client intermediary isn't strictly necessary for generated `resource.Object` implementations (though it may still be necessary for hand-written ones that use custom codecs), but is still favored because of the above two reasons.

IfSentient (approved):
Looks good, thank you! Just a few small comments

IfSentient (approved):

PR #869, opened 2025-07-16
[docs] Add migration doc for v0.40.0

IfSentient:
Add a migration guide for `v0.40.0` to migrate CUE for codegen.

### DIFF
--- a/k8s/apiserver/storage.go
+++ b/k8s/apiserver/storage.go
@@ -76,56 +76,56 @@ func matchKind(label labels.Selector, field fields.Selector) storage.SelectionPr
 	}
 }
 
-func newRegistryStatusStoreForKind(scheme *runtime.Scheme, kind resource.Kind, specStore *genericregistry.Store) *statusREST {
+func newRegistryStatusStoreForKind(scheme *runtime.Scheme, kind resource.Kind, specStore *genericregistry.Store) *StatusREST {
 	strategy := newStatusStrategy(scheme, kind.GroupVersionKind().GroupVersion())
 	return newStatusREST(specStore, strategy)
 }
 
 // NewStatusREST makes a RESTStorage for status that has more limited options.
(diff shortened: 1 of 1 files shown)


docs/migrations/v0.40.md
```diff
@@ -0,0 +1,381 @@
...
+    versions: {
+        "v1": {
+            schema: {
+                spec: {
+                    foz: int
+                }
+            }
+        }
+        "v2": {
+            validation: operations: ["CREATE","UPDATE]
```
konsalex, line 108: nit
```suggestion
            validation: operations: ["CREATE","UPDATE"]
```

docs/migrations/v0.40.md
```diff
@@ -0,0 +1,381 @@
...
+    validation: operations: ["CREATE","UPDATE"]
+}
+
+foov1: fooKind & { // join the kind data into this new object which contains the schema for the version
+    schema: {
+        spec: {
+            foo: string
+        }
+    }
+}
```
konsalex, line 145: Maybe its only for the demo what you did or I miss context? If I get this right, those fields should be emitted from the manifest generation or they are skipped even if they appear? 

I would expect something like the following 🤔 
```
#fooKind: {
	kind:       "Foo"
	plural:     "multifoo"
	conversion: true
	validation: operations: ["CREATE", "UPDATE"]
    ...
}

#foov1: #fooKind & {
	schema: {
		spec: {
			foo: string
		}
	}
}
```

konsalex (approved):
From my side seems solid, only nit comments.

I can also run a demo upgrade it in an existing codebase to validate the steps

docs/migrations/v0.40.md
IfSentient: The codegen will only pull the `manifest` key from your CUE (unless you change it with a CLI flag), so you don't need to make the other fields definitions.

Issue #460, opened 2024-11-05
Recursive Types


IfSentient:
Because [CRDs cannot handle recursive types natively](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation), we do not currently support recursive codegen from the CLI ("do not currently support" in this case means "CLI freezes up when CUE attempts to convert the recursive CUE type to OpenAPI). However, there are definitely use-cases for recursive types within app platform, so we should evaluate if we want to support recursive types in any possible case. Some questions:
* If we support recursive types in any way, how does this impact the app manifest, which inlines the schemas?
* Do we want to extend apiextensions to handle recursive types?
* Do we want to only generate _code_ for the recursive types, and use `kubernetes-preserve-unknown-fields: true` on an empty object when generating the schema (as suggested in [this kubernetes issue](https://github.com/kubernetes-sigs/controller-tools/issues/585)? We could use webhook validation to check that the actual data is correct on create/update.
* How does this impact other tooling and/or service discovery?

radiohead:
> However, there are definitely use-cases for recursive types within app platform

Do we have any real requests / questions / examples?

> Do we want to only generate code for the recursive types, and use kubernetes-preserve-unknown-fields: true on an empty object when generating the schema (as suggested in https://github.com/kubernetes-sigs/controller-tools/issues/585?

I would prefer that for a V1 / V0 approach. We could start there and investigate changing apiextensions code to support recursive types natively later.

PR #876, opened 2025-07-18
Use cog for CRD/AppManifest OpenAPI Generation

IfSentient:
This PR switches to using [grafana/cog](https://github.com/grafana/cog) instead of the CUE SDK for OpenAPI generation to use in CRDs and the AppManifest. This both unifies our code generation tooling to only using cog, as well as solves the problem with recursive CUE definitions causing the `generate` command to hang and eventually crash due to the CUE SDK's OpenAPI generation recursing indefinitely (cog's intermediate AST allows for the OpenAPI generation to work properly with recursion).

The OpenAPI returned by cog contains a list of definitions, so this is condensed into a single CRD-compatible OpenAPI schema, with recursive elements turned into objects with `x-kubernetes-preserve-additional-properties` set to `true`.

As we now support more than CRDs with the OpenAPI, the next step after this PR is to move the App Manifest's version's kind's schemas to use a full OpenAPI components set like what is generated by cog (which can properly represent the recursion by using definition references, which are unsupported by CRDs), and move the components-to-single-CRD-compatible-object logic into the AppManifest to allow the schema to be fetched as either a CRD-compatible schema or one fit to use in an API server or other machinery which can handle definition references. This was left out of this PR as this PR is designed to keep the current behavior while switching the OpenAPI generator.

Resolves https://github.com/grafana/grafana-app-sdk/issues/460


codegen/jennies/crd.go
```diff
@@ -164,143 +168,230 @@ type customResourceDefinitionMetadata struct {
...
-		typeString, ok := typ.(string)
-		if !ok {
-			// "type" property is not a string, not valid openAPI
-			return
+
+		// Create a new visited map for this branch to avoid false positives in parallel branches
+		branchVisited := make(map[string]bool)
+		for k, v := range visited {
+			branchVisited[k] = v
 		}
```
radiohead, line 265: NIT – can use `maps.Copy` here as well.

codegen/testing/golden_generated/crd/customkind.customapp.ext.grafana.com.json.txt
```diff
@@ -101,28 +100,24 @@
...
                                         "default": "default",
                                         "enum": [
-                                            "default",
+                                            "val1",
                                             "val2",
                                             "val3",
                                             "val4",
-                                            "val1"
-                                        ],
-                                        "type": "string"
```
radiohead, line 110: BUG – looks like this broke after switching to `cog` generation. We either need to patch in some sort of fix on our end (i.e. during post-processing set the type manually if it's an enum) or wait till it's fixed in `cog`.

CC @spinillos 

This will currently not work for CRDs:
```console
❯ k --context=dev-us-central-0-app-platform apply -f test-enum-crd.yaml
The CustomResourceDefinition "testenums.example.com" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[status].type: Required value: must not be empty for specified object fields
```

With CRD being:
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: testenums.example.com
spec:
  group: example.com
  names:
    kind: TestEnum
    listKind: TestEnumList
    plural: testenums
    singular: testenum
  scope: Namespaced
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                status:
                  enum:
                    - "pending"
                    - "running"
                    - "completed"
                    - "failed"
                  default: "pending"
              required:
                - status
          required:
            - spec 
```

codegen/jennies/crd.go
```diff
@@ -164,143 +168,230 @@ type customResourceDefinitionMetadata struct {
...
+		Enum:                 schema.Value.Enum,
+		Title:                schema.Value.Title,
+		AdditionalProperties: schema.Value.AdditionalProperties,
+		Nullable:             schema.Value.Nullable,
+		ReadOnly:             schema.Value.ReadOnly,
+		WriteOnly:            schema.Value.WriteOnly,
+		AllOf:                make([]*openapi3.SchemaRef, 0),
+		OneOf:                make([]*openapi3.SchemaRef, 0),
+		AnyOf:                make([]*openapi3.SchemaRef, 0),
 	}
```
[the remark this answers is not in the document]
IfSentient: Good catch, we probably do (even though I don't think the original will have extensions in our particular use-case, it may in the future).

radiohead (approved):

PR #879, opened 2025-07-22
Add Custom Subresource Request Method to Clients

IfSentient:
Now that arbitrary custom subresources are supported for API server apps, we need to update the `resource.Client` (and other clients built on top of it) to allow for making arbitrary subresource requests. This PR introduces a new method:
```go
SubresourceRequest(context.Context, resource.Identifier, resource.CustomRouteRequestOptions) ([]byte, error)
```
to `resource.Client` (and similar methods to `resource.TypedClient` and `resource.NamespacedClient`), and implements it in `k8s.Client`. The underlying implementation in `k8s.gvclient` can also be used for custom resource routes once implemented.

Another improvement to be made in a subsequent PR will be to generate go clients for each kind that include dedicated, strongly-typed methods for handling each custom subresource.

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -239,6 +239,16 @@ func (c *Client) Watch(ctx context.Context, namespace string, options resource.W
 	return c.client.watch(ctx, namespace, c.schema.Plural(), c.schema.ZeroValue(), options, c.codec)
 }
 
+// SubresourceRequest makes a request to an arbitrary subresource of the object identified by identifier.
+// Verb and Path should be included in the options.
+func (c *Client) SubresourceRequest(ctx context.Context, identifier resource.Identifier, opts resource.CustomRouteRequestOptions) ([]byte, error) {
+	if c.schema.Scope() == resource.ClusterScope && identifier.Namespace != resource.NamespaceAll {
+		return nil, fmt.Errorf("cannot watch resources with schema scope \"%s\" in namespace \"%s\", must be NamespaceAll (\"%s\")",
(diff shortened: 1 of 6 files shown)


k8s/client.go
```diff
@@ -239,6 +239,16 @@ func (c *Client) Watch(ctx context.Context, namespace string, options resource.W
 	return c.client.watch(ctx, namespace, c.schema.Plural(), c.schema.ZeroValue(), options, c.codec)
 }
 
+// SubresourceRequest makes a request to an arbitrary subresource of the object identified by identifier.
+// Verb and Path should be included in the options.
+func (c *Client) SubresourceRequest(ctx context.Context, identifier resource.Identifier, opts resource.CustomRouteRequestOptions) ([]byte, error) {
```
ferruvich, line 244: Should we also validate the correctness of `opts`? 
i.e. existing Verb/Path

resource/client.go
```diff
@@ -158,6 +161,20 @@ type WatchEvent struct {
 	Object Object
 }
 
+// CustomRouteRequestOptions contains the options for a custom route request
+type CustomRouteRequestOptions struct {
+	// Path is the path of the custom route, from the base of the resource or namespace/api version
+	Path string
```
ferruvich, line 167: Do you think we should make an example here, to better explain how a path should be?

ferruvich (approved):
LGTM! Left one tiny comment

radiohead (approved):

PR #890, opened 2025-07-29
[codegen] Generate Typed Clients for Resources

IfSentient:
With the addition of custom subresource routes, the generic way of querying for a subresource using `resource.Client.SubresourceRequest` cannot be typed in a useful way, and will always consume a generic body and return generic bytes. To help make it simpler to use custom subresources, this PR introduces generated, typed clients for each kind that include methods for each custom subresource that accept strongly typed parameters and return a typed body response.

This PR also changes the associator functions to use a single interface rather than adding more functions to `apiserver.NewDefaultInstaller`.


ferruvich:
@IfSentient sorry for the huge late on this - could you please rebase it? I'll have a look at it tomorrow morning 👍🏼

ferruvich (approved):

PR #895, opened 2025-07-31
Add EventTimeout option and use contextual `cache.ListWatch` methods

radiohead:
### What

- **Added `EventTimeout` option** to `CustomCacheInformerOptions` and `KubernetesBasedInformerOptions`
- **Updated `ListerWatcher`** to use contextual `ListWithContextFunc` and `WatchFuncWithContext`
- **Modified `NewCustomCacheInformer` signature** to pass options via `CustomCacheInformerOptions` struct

### Why

- **EventTimeout prevents resource leaks** by dropping events that take too long to process
- **Contextual ListerWatcher improves cancellation handling** and resource cleanup during watch operations
- **Configurable options provide better control** over informer behavior without breaking existing functionality

### API Changes

⚠️ **Breaking Change**: `NewCustomCacheInformer` now requires a `CustomCacheInformerOptions` parameter:

```golang
// Before
inf := NewCustomCacheInformer(store, lw, kind)

// After  
inf := NewCustomCacheInformer(store, lw, kind, CustomCacheInformerOptions{})
```

### DIFF
--- a/codegen/templates/app/app.tmpl
+++ b/codegen/templates/app/app.tmpl
@@ -1,11 +1,9 @@
 package {{.PackageName}}
 
 import (
-    "context"
     "fmt"
 
 	"github.com/grafana/grafana-app-sdk/app"
-	"github.com/grafana/grafana-app-sdk/logging"
 	"github.com/grafana/grafana-app-sdk/simple"
 
 	generated "{{.Repo}}/{{.CodegenPath}}"
@@ -28,12 +26,6 @@ func New(cfg app.Config) (app.App, error) {
 	config := simple.AppConfig{
 		Name:           "{{.ProjectName}}",
 		KubeConfig:     cfg.KubeConfig,
-		InformerConfig: simple.AppInformerConfig{
-		    ErrorHandler: func(ctx context.Context, err error) {
-                // FIXME: add your own error handling here
-                logging.FromContext(ctx).With("error",err).Error("Informer processing error")
-            },
-		},
 		ManagedKinds: []simple.AppManagedKind{ {{ range $key, $val := .GVToKindAll }}{{ range $val }}
 		    {
 		        Kind: {{if $.KindsAreGrouped }}{{$.ToPackageNameVariable ($key.String)}}.{{.Kind}}Kind(){{else}}{{ $.ToPackageName .MachineName }}{{ $.ToPackageName $key.Version}}.Kind(){{end}}, {{ if eq $key.Version .Current }}
--- a/operator/informer_kubernetes.go
+++ b/operator/informer_kubernetes.go
@@ -30,7 +30,7 @@ type KubernetesBasedInformer struct {
 	schema                resource.Kind
 	runContext            context.Context
 	eventTimeout          time.Duration
-	errorHandlerFn        func(context.Context, error)
+	errorHandler          func(context.Context, error)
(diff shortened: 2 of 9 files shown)


operator/informer_kubernetes.go
```diff
@@ -22,24 +22,38 @@ var (
...
 }
 
-type KubernetesBasedInformerOptions struct {
+// InformerOptions are generic options for all Informer implementations.
+type InformerOptions struct {
 	// ListWatchOptions are the options for filtering the watch based on namespace and other compatible filters.
 	ListWatchOptions ListWatchOptions
 	// CacheResyncInterval is the interval at which the informer will emit CacheResync events for all resources in the cache.
 	// This is distinct from a full resync, as no information is fetched from the API server.
-	// An empty value will disable cache resyncs.
```
IfSentient, line 41: I think we should still allow `CacheResyncInterval` being zero to disable cache resync. It looks like we still do, but this line of documentation is replaced.

operator/informer_kubernetes.go
```diff
@@ -22,24 +22,38 @@ var (
...
+	// Changes to this value after run() is called will not take effect.
```
IfSentient, line 43: NIT: `run()` -> `Run()`

IfSentient (approved):
LGTM, a few minor tweaks that I think should go in before merging.

codegen/templates/app/app.tmpl
```diff
@@ -28,12 +26,6 @@ func New(cfg app.Config) (app.App, error) {
 	config := simple.AppConfig{
 		Name:           "{{.ProjectName}}",
 		KubeConfig:     cfg.KubeConfig,
-		InformerConfig: simple.AppInformerConfig{
```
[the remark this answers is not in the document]
radiohead: > We should still provide sample code with the error handler when generating

Should we? I see a bunch of apps where the error handler is left as-is from the generated code and IMO it's questionable in terms of extra value.

I am happy to put this back but I don't see any benefit in shipping something that users don't customize (and we have a default error handler already).

operator/informer_kubernetes.go
radiohead: Fixed!

operator/informer_kubernetes.go
radiohead: Fixed!

spinillos (approved):

codegen/templates/app/app.tmpl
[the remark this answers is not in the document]
radiohead: Let's remove it then!

operator/informer_kubernetes.go
```diff
@@ -70,9 +94,15 @@ func NewKubernetesBasedInformer(sch resource.Kind, client ListWatchClient, optio
 		healthCheckName = fmt.Sprintf("%s?%s", healthCheckName, strings.Join(params, "&"))
 	}
 
+	errorHandler := DefaultErrorHandler
+	if options.ErrorHandler != nil {
+		errorHandler = DefaultErrorHandler
```
spinillos, line 99: This should be `options.ErrorHandler` 😅 

```suggestion
		errorHandler = options.ErrorHandler
```

radiohead: You didn't see anything 🙈

PR #945, opened 2025-08-29
[apiserver] Custom Resource Routes

IfSentient:
Now that the `k8s/apiserver` supports custom subresource routes, we need to enahnce this functionality with the ability to specify custom routes at the resource level (e.g. `example.grafana.app/v1alpha1/foo` or `example.grafana.app/v1alpha1/namespaces/default/foo`, where `foo` is the custom route and not a kind). 

This PR introduces this functionality and has a few slight updates to the AppManifest to account for handling resource routes at either the namespace or cluster scope. The UX for registering these route handlers in `simple.App` feels a bit clunky, and I'd love any ideas on improving it.

The route registration itself is not done via storage at this point, as there are several limitations that we discovered while working on subresource routes (a large one being that request parameters cannot be specified per-verb, only responses), so instead the new handler is registered directly with the webservice handling the API group. This does mean that we are no longer able to accept an interface for the genericapiserver, as the handler we need to access is an exported field, and does not have a function for accessing it (we could wrap this in order to accept an interface, but it felt like a worse UX, especially since we're still heavily depending on the logic of genericapiserver anyway).

### DIFF
--- a/codegen/testing/golden_generated/manifest/go/groupbygroup/customapp_manifest.go.txt
+++ b/codegen/testing/golden_generated/manifest/go/groupbygroup/customapp_manifest.go.txt
@@ -13,6 +13,7 @@ import (
 	"github.com/grafana/grafana-app-sdk/app"
 	"github.com/grafana/grafana-app-sdk/resource"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kube-openapi/pkg/spec3"
 
 	v0_0 "codegen-tests/pkg/generated/customapp/v0_0"
 	v1_0 "codegen-tests/pkg/generated/customapp/v1_0"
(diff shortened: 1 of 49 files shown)


codegen/testing/golden_generated/manifest/go/groupbygroup/customapp_manifest.go.txt
```diff
@@ -104,8 +114,23 @@ func ManifestCustomRouteQueryAssociator(kind, version, path, verb string) (goTyp
...
 
+var customRouteToGoRequestBodyType = map[string]any{}
+
+func ManifestCustomRouteRequestBodyAssociator(kind, version, path, verb string) (goType any, exists bool) {
+	if len(path) > 0 && path[0] == '/' {
+		path = path[1:]
+	}
+	goType, exists = customRouteToGoRequestBodyType[fmt.Sprintf("%s|%s|%s|%s", version, kind, path, strings.ToUpper(verb))]
+	return goType, exists
+	return nil, false
```
spinillos, line 125: This is bad generated.

PR #972, opened 2025-09-08
Function to Auto-Retry of Update Conflicts

IfSentient:
A useful feature of the app-sdk would be to have the ability to auto-retry conflicts, as often an app author will not want to have to write specify error-based retry handling for their app logic.

This PR introduces a function in the `resource` package called `UpdateObject`, which does a 409-retryable `Put` request using a client, and accepts a mutation function rather than an object, in order to be able to use get-and-put semantics for both the initial and conflict retries.

This can be used from the stores by calling `Client()` on the store, for example (updating the status):
```go
updated, err := UpdateObject(ctx, store.Client(), func(obj **v1alpha1.MyObject, isRetry bool) {
    (*obj).Status.MyList = append((*obj).Status.MyList, "to-add")
    return nil
}, resource.UpdateOptions{
    Subresource: resource.SubresourceStatus,
})
```


resource/updater.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"errors"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// UpdateObject ensures an object is updated by calling Get on the provided client to get the current state of the object,
+// then runs updateFunc on it to update it, and finally calls UpdateInto with the updated object and provided config,
+// using the ResourceVersion of the updated object. If the update fails due to a 409/Conflict, it will retry the whole process up to two times.
+func UpdateObject[T Object](ctx context.Context, client Client, identifier Identifier, updateFunc func(obj *T, isRetry bool) error, opts UpdateOptions) (T, error) {
```
radiohead, line 13: The pointer-to-pointer semantics make this very confusing IMO. Just for that I would change the signature of the function to `(T, bool) -> (T, error)`.

radiohead (approved):
Looks good, but the function signature makes the resulting code confusing because of pointer-to-pointer usage.

radiohead (approved):

PR #1000, opened 2025-09-19
[AppManifest] Add apiVersion, kind, metadata, and spec as required fields when calling AsKubeOpenAPI

IfSentient:
Add apiVersion, kind, metadata, and spec as required fields when calling `AsKubeOpenAPI`, as these should all be required. `spec` is only added as required if present (there may be use-cases for kinds that don't have a spec).


gillesdemey (approved):
✅

PR #1023, opened 2025-10-14
[simple.App] Allow Config to Supply a ClientGenerator

IfSentient:
Allow passing a client generator in the config, rather than always generating one with the default client config. If one is not provided in the config, the default is the prior logic.


radiohead (approved):
Meant to request changes.

ferruvich (approved):

PR #1024, opened 2025-10-15
Introduce resource.PartialObject and k8s.DeferredNegotiatedSerializer…

IfSentient:
Introduce `resource.PartialObject` and `k8s.DeferredNegotiatedSerializer` to allow for deferring more expensive unmarshaling for a given kind. This is an experimental feature to see if this accelerates initial start-up processing of large numbers of objects in CPU-constrained environments, and may be removed in a future release if not effective.

Includes changes from not-yet-merged https://github.com/grafana/grafana-app-sdk/pull/1023 to allow specifying the negotiated serializer using `k8s.ClientConfig`.

### DIFF
--- a/resource/partialobject.go
+++ b/resource/partialobject.go
@@ -16,19 +16,12 @@ var _ Object = &PartialObject{}
 // PartialObject implements resource.Object but only actually contains metadata information, and the raw payload that was used for unmarshaling.
 // This is useful in accelerating the unmarshal process that is done serially with a NegotiatedSerializer in kubernetes watch requests,
 // but does consume more memory as the entire original payload is embedded to avoid needing to copy or attempt to understand the non-metadata fields.
+//
+// PartialObject is _Experimental_ and may be removed in a future release
 type PartialObject struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
 	Raw               []byte `json:"-"`
-}
-
-func NewPartialObject(raw []byte, decoder func([]byte, any) error) (*PartialObject, error) {
-	p := PartialObject{}
-	if err := decoder(raw, &p); err != nil {
-		return nil, err
-	}
-	p.Raw = raw
-	return &p, nil
 }
 
 type metadataOnlyObject struct {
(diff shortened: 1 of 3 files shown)


simple/app.go
```diff
@@ -267,10 +270,14 @@ type AppVersionRouteHandlers map[AppVersionRoute]AppCustomRouteHandler
 // AppConfig MUST contain a valid KubeConfig to be valid.
 // Watcher/Reconciler error handling, retry, and dequeue logic can be managed with AppConfig.InformerConfig.
 func NewApp(config AppConfig) (*App, error) {
+	clients := config.ClientGenerator
+	if clients == nil {
+		k8s.NewClientRegistry(config.KubeConfig, k8s.DefaultClientConfig())
```
radiohead:
```suggestion
		clients = k8s.NewClientRegistry(config.KubeConfig, k8s.DefaultClientConfig())
```

resource/partialobject.go
```diff
@@ -0,0 +1,170 @@
...
+	*metav1.ObjectMeta `json:"metadata"`
+}
+
+func (p *PartialObject) UnmarshalJSON(b []byte) error {
+	md := metadataOnlyObject{}
+	if err := json.Unmarshal(b, &md); err != nil {
+		return err
+	}
+	p.TypeMeta = *md.TypeMeta
+	p.ObjectMeta = *md.ObjectMeta
```
radiohead: These could be nil, couldn't they?

radiohead (commented):
A couple of minor comments, but also, could we add a benchmark? Something like:
```go
// resource/partialobject_test.go
func BenchmarkPartialObjectUnmarshal(b *testing.B) {
    data := []byte(`{"apiVersion":"v1","kind":"Pod","metadata":{"name":"test","namespace":"default"},"spec":{"containers":[...]}}`)
    for i := 0; i < b.N; i++ {
        var obj PartialObject
        _ = json.Unmarshal(data, &obj)
    }
}

func BenchmarkTypedObjectUnmarshal(b *testing.B) {
    data := []byte(`{"apiVersion":"v1","kind":"Pod","metadata":{"name":"test","namespace":"default"},"spec":{"containers":[...]}}`)
    for i := 0; i < b.N; i++ {
        var obj TypedSpecObject[PodSpec]
        _ = json.Unmarshal(data, &obj)
    }
}
```

radiohead:
@IfSentient this didn't get any traction, should we drop it?

Issue #959, opened 2025-09-03
[apiserver] Use Defaults for Fields on Create


IfSentient:
When creating a new object in an apiserver, while defaults are shown in the OpenAPI for the kind, they are not used in the actual create operation.

This seems to stem from a lack of awareness of the OpenAPI when creating the resource (create handler is [installed here](https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/installer.go#L920) and the [handler doesn't use any OpenAPI](https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/handlers/create.go#L53)). A possible solution is to return an object with the defaults from the storage's `New()` function, [as it appears that is used to generate the "template" object which is combined with the incoming one in the create handler](https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/handlers/create.go#L195-L199). We will likely want to generate code for a new object with the defaults as specified by the CUE, rather than working it out at runtime.

PR #1103, opened 2025-11-18
[apiserver] Use Defaults for Fields on Create

amalavet:
## What Changed? Why?
When creating a new object in an apiserver, while defaults are shown in the OpenAPI for the kind, they are not used in the actual create operation. To address this, I modified codegen to include a `New` function for every kind which recurses through the schema and applies all defaults.

Addresses Issue #959 

### How was it tested?
1. Defined a schema with nested defaults for the example apiserver `TestKind`
2. Generated the new code with the changes in the PR
3. Created a new object and verified that the schema is populated with the defaults (see below

```bash
curl -k -X POST -H "content-type: application/json" -d '{"apiVersion":"example.ext.grafana.com/v1alpha1","kind":"TestKind","metadata":{"name":"myObj","namespace":"default"},"spec":{}}' https://127.0.0.1:6443/apis/example.ext.grafana.com/v1alpha1/namespaces/default/testkinds
```
```json
{
  "kind": "TestKind",
  "apiVersion": "example.ext.grafana.com/v1alpha1",
  "metadata": {
    "name": "myObj",
    "namespace": "default",
    "uid": "9a901cc9-dc79-4612-91db-bffeece329ce",
    "resourceVersion": "15",
    "creationTimestamp": "2025-11-18T23:56:00Z",
    "managedFields": [
      {
        "manager": "curl",
        "operation": "Update",
        "apiVersion": "example.ext.grafana.com/v1alpha1",
        "time": "2025-11-18T23:56:00Z",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:spec": {
            "f:foo": {
              "f:bar": {
                "f:baz": {
                  "f:value": {}
                },
                "f:value": {}
              },
              "f:foo": {}
            },
            "f:testField": {}
          }
        }
      }
    ]
  },
  "spec": {
    "testField": "default value",
    "foo": {
      "foo": "foo",
      "bar": {
        "value": "bar",
        "baz": {
          "value": 10
        }
      }
    }
  },
  "status": {},
  "mysubresource": {
    "extraValue": ""
  }
}
```

### DIFF
--- a/codegen/cuekind/def.cue
+++ b/codegen/cuekind/def.cue
@@ -94,6 +94,10 @@ SchemaWithOperatorState: Schema & {
     // responseMetadata allows codegen to include kubernetes metadata in the generated response object.
     // It is also copied into the AppManifest responseMetadata for use in kube-OpenAPI generation.
     responseMetadata: #CustomRouteResponseMetadata
+    // extensions are all openAPI extensions that you wish to apply to this route.
+    extensions: {
+    	[=~"^x-(.+)$"]: _
+    }
 }
 #CustomRoutePath: string
 #CustomRouteMethod: "GET" | "POST" | "PUT" | "DELETE" | "PATCH" | "*"
--- a/codegen/cuekind/parser.go
+++ b/codegen/cuekind/parser.go
@@ -561,6 +561,16 @@ func (*Parser) parseCustomRoutes(customRoutesVal cue.Value) (map[string]map[stri
 			if nameStrVal.Exists() && !nameStrVal.IsNull() {
 				route.Name, _ = nameStrVal.String()
 			}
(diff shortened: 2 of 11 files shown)


amalavet:
> I'll take a closer look later today, but off the top of my head I wonder if we should instead use the k8s defaulter interface approach?
> 
> Here's a quick llm-generated explainer of defaulting, based on the `kubernetes/kubernetes` codebase – https://gist.github.com/radiohead/da40d975d9a793dcc01b34a204ccf4e0.

I took a look at this. Because both validation and defaults are applied **after** decoding, there's no way to determine if the client sent `{"spec":{"value:""}}` v.s. `{"spec":{}}`. More specifically, when decoding into a go type, unknown fields to be their zero types (i.e. 0, "" etc), meaning that both `{"spec":{"value:""}}` and `{"spec":{}}` result in a go struct like `{ Spec: {Value:""}}`. As a result, we i

Issue #848, opened 2025-06-26
bug: can't use "go" in --gogenpath of grafana-app-sdk generate


joeblubaugh:
This works:
```
> grafana-app-sdk generate -g my-custom-path/golang
```

This fails:
```
> grafana-app-sdk generate -g my-custom-path/go
Error: 1 error occurred:
        * ManifestGoGenerator: 6:9: expected 'IDENT', found 'go' (and 1 more errors) for input "kinds"
```

joeblubaugh:
A related argument parsing issue (let me know if you want a separate bug):
```
❯ grafana-app-sdk generate --source kinds/service-model --gogenpath "types/golang/service-model"
Error: 1 error occurred:
        * ManifestGoGenerator: 6:16: expected ';', found '-' for input "servicemodel"
```

amalavet:
Here's a possible approach, not sure the downstream impact of this but I think it would be safe: https://github.com/grafana/grafana-app-sdk/pull/1105/files

PR #1105, opened 2025-11-19
[codegen] sanitize package names prior to manifest generation

amalavet:
## What Changed? Why?

Modify `templates.ToPackageName` to handle sanitizing packages that include certain go keywords, or invalid characters.

Addresses Issue #848

### How was it tested?
Successful local package generation for both examples provided in the the issue

### Notes to Reviewers
Please let me know if there could be any downstream negative effects of this change to the `templates.ToPackageName` func.

### DIFF
--- a/codegen/templates/templates.go
+++ b/codegen/templates/templates.go
@@ -627,18 +627,41 @@ func WriteRuntimeObjectWrapper(metadata RuntimeObjectWrapperMetadata, out io.Wri
 	return templateRuntimeObject.Execute(out, metadata)
 }
 
+// goKeywords is the set of Go reserved keywords that cannot be used as package names
+var goKeywords = map[string]struct{}{
+	"break":       {},
+	"case":        {},
+	"chan":        {},
+	"const":       {},
+	"continue":    {},
+	"default":     {},
+	"defer":       {},
+	"else":        {},
+	"fallthrough": {},
+	"for":         {},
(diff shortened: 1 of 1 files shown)


codegen/templates/templates.go
```diff
@@ -627,8 +627,41 @@ func WriteRuntimeObjectWrapper(metadata RuntimeObjectWrapperMetadata, out io.Wri
 	return templateRuntimeObject.Execute(out, metadata)
 }
 
+// goKeywords is the set of Go reserved keywords that cannot be used as package names
+var goKeywords = map[string]struct{}{
```
Proximyst: why not just a slice + `slices.Contains`? more concise & compact and I can't imagine it's a measurable perf impact :)

codegen/templates/templates.go
```diff
@@ -627,8 +627,41 @@ func WriteRuntimeObjectWrapper(metadata RuntimeObjectWrapperMetadata, out io.Wri
...
+	"select":      {},
+	"struct":      {},
+	"switch":      {},
+	"type":        {},
+	"var":         {},
+}
+
 // ToPackageName sanitizes an input into a deterministic allowed go package name.
 // It is used to turn kind names or versions into package names when performing go code generation.
 func ToPackageName(input string) string {
```
Proximyst: do package names permit uppercase?

PR #1118, opened 2025-11-24
[codegen] fix unused fmt bug

amalavet:
## What Changed? Why?
Sometimes fmt is not used, this checks the only condition it would be used.

### How was it tested?
Locally

### DIFF
--- a/codegen/templates/schema.tmpl
+++ b/codegen/templates/schema.tmpl
@@ -7,10 +7,8 @@ package {{.Package}}
 {{$needsFmt := false}}{{range .SelectableFields}}{{if ne .Type "string"}}{{$needsFmt = true}}{{end}}{{end}}
 import (
     {{if gt $sfl 0}}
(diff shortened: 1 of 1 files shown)


codegen/templates/schema.tmpl
```diff
@@ -4,10 +4,13 @@
 
 package {{.Package}}
 {{$sfl := len .SelectableFields}}
+{{$needsFmt := false}}{{range .SelectableFields}}{{if ne .Type "string"}}{{$needsFmt = true}}{{end}}{{end}}
 import (
     {{if gt $sfl 0}}
-    "fmt"
     "errors"
+    {{if $needsFmt}}
```
IfSentient: Can we change this to 
```
"errors"{{if $needsFmt}}
"fmt"{{end}}
```
So that we don't end up with extra line breaks that will upset `gofmt`?

PR #1122, opened 2025-11-25
[apiserver] Allow API Installer to install APIVersions with no kinds

IfSentient:
## What Changed? Why?

Currently, attempting to install an app with no kinds, but custom resource routes, result sin an error when setting the version priority. This PR fixes the underlying problem that an app without kinds doesn't get a webservice installer, even if it has routes for a version, so doesn't get routes added. This PR also removes the "silent" failure mode of failing to install routes because the webservice to install them on cannot be found, and instead returns an error for this.

### How was it tested?

Locally, on the example apiserver.

N/A - bugfix

### DIFF
--- a/examples/apiserver/apis/example/v0alpha1/testkind_client_gen.go
+++ b/examples/apiserver/apis/example/v0alpha1/testkind_client_gen.go
@@ -3,8 +3,9 @@ package v0alpha1
 import (
 	"context"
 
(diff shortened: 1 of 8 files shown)


stevesg:
Ship it :shipit:

stevesg (approved):

owensmallwood (approved):

PR #1159, opened 2026-01-05
Update `k8s.io` dependencies to `1.35.0`

IfSentient:
## What Changed? Why?

Updated the `k8s.io` dependencies to `1.35.0`. This change contained breaking changes to the API Server OpenAPI behavior, which required extensive changes to the codegen, app manifest `AsKubeOpenAPI` logic, and `apiserver.defaultInstaller`'s `GetOpenAPIDefinitions` and `AddToScheme`.

The meat of the change that required changes in the app-sdk code was that kubernetes no longer uses the reflected package path as the OpenAPI key, modified to use dots instead of slashes. Now it relies on a new interface, [util.OpenAPIModelNamer](https://pkg.go.dev/k8s.io/kube-openapi@v0.0.0-20251125145642-4e65d59e963e/pkg/util#OpenAPIModelNamer) to match types in the `runtime.Scheme` up to the fields in the OpenAPI. As a side-effect of this, since it no longer corrects the slash-separated reflected paths when ingesting the OpenAPI schema, `fieldManager` cannot associate types which use slash-separation, as `jsonpointer` translates a [forward slash to ~1](https://github.com/go-openapi/jsonpointer/blob/master/pointer.go#L553), and the kubernetes code does not when `fieldManager` does its lookups (resulting in it attempting to lookup `a~1b~1c.Foo` when the reference OpenAPI name is `a/b/c.Foo`). 

The fix for this was to implement `OpenAPIModelNamer` for all generated go types. This has been done in a somewhat-hacky way by regex matching on `type <X> struct` and adding an `OpenAPIModelName` function for `X` at the end of the file. This likely should be fixed more permanently in `grafana/cog` itself. 

The generated name needed to be in the kubernetes-expected format of dot-separated general-to-specific (e.g. `com.github.grafana` instead of `github.com.grafana`) paths ending in the type name, and the `apiserver` defaultInstaller's needed updating to ensure that all the definition paths and names properly matched.

Unrelated to the OpenAPI changes, the feature flag `WatchListClient` defaults to enabled as of `1.35.0`, and the code for the Reflector uses an in-memory temporary store to cache all the "initial list" (watch with initial events) items before syncing to the actual store. This results in CustomCacheInformer not saving any memory compared to the default. Since the logic that uses the temporary in-memory store is only used for consistency checking against `LIST` for k8s integration testing, this logic was removed in this PR and the watchlist adds directly to the provided store. This has the added benefit of reducing start-up time, as all events don't need to be streamed into memory before processing of them can begin (as before the events wouldn't be processed until `Replace()` was called on the store after all events were streamed).

### How was it tested?

Tested the API server using the `examples/apiserver` apiserver, as well as a separate test project. 

Remaining testing is the unit testing and integration tests run by github actions.

No additional documentation except in some comments in jennies for TODOs or reasoning for the changes. Customers should not see any breaking changes aside from the names of OpenAPI references changing in the `/openapi` endpoints from the API server, and some generated go types changing names (shared types for custom routes are now named more correctly without exceedingly long names due to not using the type prefix).


codegen/jennies/gotypes.go
```diff
@@ -235,6 +267,28 @@ func GoTypesFromCUE(v cue.Value, cfg CUEGoConfig, maxNamingDepth int) ([]byte, e
...
+	return "%s"
+}
+`, string(matches[1]), openAPIName))
+			if err != nil {
+				return nil, fmt.Errorf("failed to write OpenAPIModelName() function: %w", err)
+			}
+		}
+		files[0].Data = append(files[0].Data, appendBytes.Bytes()...)
+	}
+
```
amalavet: @IfSentient 
Here is a fix for this using cog: https://github.com/grafana/grafana-app-sdk/commit/1dbbcbceb6606f44b76d27ce860896c0227e8351
It relies on a branch of cog I created here: https://github.com/grafana/cog/pull/903

radiohead (approved):

PR #1306, opened 2026-03-19
Add prometheus metrics for simple.App admission chain

IfSentient:
## What Changed? Why?

Added prometheus metrics to instrument the admission chain in `simple.App` (Validate, Mutate, Convert, and CallCustomRoute methods). This enables operators to observe and monitor the performance of their admission logic.

### How was it tested?

Existing tests in `simple/` pass. The metrics follow the same patterns used by `operator.InformerController` and `k8s.ClientRegistry`, using the standard `metrics.Config` for namespace and native histogram settings.

Metrics are configured via `AppConfig.MetricsConfig` and exposed through the existing `PrometheusCollectors()` method. Three metric groups (admission, conversion, custom_route) each expose latency histograms, request counters, and in-flight gauges.

### Notes to Reviewers

- In-flight gauges increment before handler execution and decrement after (via inline Inc/Dec, not defer)
- Metrics only recorded when handlers are actually invoked (ErrNotImplemented/ErrCustomRouteNotFound do not record)
- Label cardinality follows existing patterns: admission uses operation/action labels, conversion uses source/target version, custom routes use method/path


simple/app.go
```diff
@@ -320,6 +341,80 @@ func NewApp(config AppConfig) (*App, error) {
 		cfg:                config,
 		collectors:         make([]prometheus.Collector, 0),
 	}
+	// Create admission chain metrics
+	mcfg := config.MetricsConfig
+	a.admissionLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:                       mcfg.Namespace,
```
radiohead: I know we use namespace from metrics config across the board, but when working on a shared dashboard for app-sdk operators I've realized that this is a major blocker - because different services have different metric names.

Should we disallow namespace customization and use a hard-coded `grafana_app_` namespace prefix?

radiohead (dismissed):
LGTM, left a question and a minor suggestion.

simple/app.go
IfSentient: Sorry about the late reply on this, this PR totally slipped my radar. Perhaps we default the namespace to `grafana_app_`, but still allow for customization in cases where it's needed? Probably should be a separate PR to put this behavior in for all places we use the metric config.

amalavet (approved):

IfSentient:
replaced by https://github.com/grafana/grafana-app-sdk/pull/1572

PR #1323, opened 2026-04-07
Improve o11y coverage for reconcilers

wbrowne:
## What Changed? Why?

The reconciliation pipeline had two observability gaps:
- Retry processor itself is a bit of a black box with no metrics. You can't tell how many retries are pending, how long they wait, or what proportion are succeeding vs exhausting their policy.
- Before retry queuing, errors only become visible at retry_processor_enqueued_total, which misses the first failure entirely if a retry succeeds, and gives no signal when retries are disabled.

This PR adds: 
  - reconcilerErrors and watcherErrors count failures at the point they occur before any retry logic so you get an accurate error rate regardless of retry behaviour. Alerting on these catches regressions immediately rather than waiting for retries to exhaust.  
  - The retry processor metrics (enqueued_total, executions_total, pending_total, queue_wait_duration_seconds) give you the full retry lifecycle: how much work is queued, how long it's delayed, and whether it's ultimately succeeding, being explicitly requeued, retrying on error, or failing permanently. 

`operator/retry_processor.go`
  - MetricsConfig metrics.Config on RetryProcessorConfig (inherits from informer so all related reconciler metrics share the same namespace)
  - 4 Prometheus metrics: enqueued_total, executions_total (action+result labels), pending_total (GaugeFunc), queue_wait_duration_seconds 

`operator/informer_controller.go`
  - reconcilerErrors *prometheus.CounterVec — incremented on first reconciler error, before retry queuing
  - watcherErrors *prometheus.CounterVec — incremented in each watcher closure (add/update/delete) on error

The PR also ensures all metrics as exposed via metrics.Provider in simple.App. This makes it possible for the caller to register the metrics.

### How was it tested?

- Manually tested locally
- Added unit tests

N/A (AFAIK 😅)


wbrowne:
@codex review

simple/app.go
```diff
@@ -544,7 +555,10 @@ func (a *App) RegisterKindConverter(groupKind schema.GroupKind, converter k8s.Co
 
 // PrometheusCollectors implements metrics.Provider and returns prometheus collectors used by the app for exposing metrics
 func (a *App) PrometheusCollectors() []prometheus.Collector {
-	return a.collectors
+	collectors := make([]prometheus.Collector, 0, len(a.collectors))
+	collectors = append(collectors, a.collectors...)
+	collectors = append(collectors, a.informerController.PrometheusCollectors()...)
+	return collectors
```
amalavet: Capacity hint only covers `a.collectors`, so the second `append` reallocates.

Either just do:
```suggestion
return append(a.collectors, a.informerController.PrometheusCollectors()...)
```

Or if you want to preallocate, do something like:
```go
ctrl := a.informerController.PrometheusCollectors()
out := make([]prometheus.Collector, 0, len(a.collectors)+len(ctrl))
return append(append(out, a.collectors...), ctrl...)
```

simple/app.go
```diff
@@ -114,6 +115,8 @@ type App struct {
 type AppConfig struct {
 	Name       string
 	KubeConfig rest.Config
+	// MetricsConfig controls the Prometheus namespace and histogram settings for all metrics produced by this app.
+	MetricsConfig metrics.Config
```
amalavet: `MetricsConfig` is only consumed by the informer controller today, but sits at the top of `AppConfig` which suggests app-wide scope. Moving it under `AppInformerConfig` keeps the config grouped with its only consumer and avoids signaling a broader contract than currently exists.
```suggestion
```

amalavet (approved):
Thanks for the addition!

PR #1349, opened 2026-04-15
feat: route DynamicPatcher per-group via ClientGenerator

amalavet:
## What Changed? Why?

`DynamicPatcher` used one `rest.Config` for all groups, so finalizer patches hit the wrong apiserver when `KubeConfigProvider` routes groups to different hosts. It also stored subresource entries over parent entries in its preferred-version cache, producing broken request URLs.

Motivating error seen in a multi-apiserver deployment of the example todo app:

```
{"level":"ERROR","msg":"error adding in-progress finalizer","component":"OpinionatedWatcher","kind":"Todo","namespace":"default","name":"todo-l75hd","error":"preferred resource not found for kind 'Todo.exampletodoapp.ext.grafana.app'"}
```

Discovery was hitting the local apiserver, which doesn't know about `exampletodoapp.ext.grafana.app` — that group lives on a different host. The cache never populated for it, so every finalizer patch failed with "preferred resource not found".

Key changes:

- **Per-group routing.** `DynamicPatcher` now takes a `ClientGenerator`; discovery goes through `ClientGenerator.DiscoveryClient()` and patch/get through `ClientFor` at the preferred version. Per-group `KubeConfigProvider` routing is preserved end-to-end.
- **Subresource filter.** `APIGroupInfo` skips entries whose `Name` contains `/` (`pods/status`, `deployments/scale`, etc.) so they don't overwrite the parent resource in the cache.
- **Scoped cache state.** `preferred` and `lastUpdate` are keyed per group; one group's discovery failure no longer poisons the whole cache. Target-group failures with no usable results surface a real error instead of the downstream "preferred resource not found".
- **`updateInterval` fix.** `>= 0` → `> 0`, so `interval == 0` actually disables refresh as the docstring claims.

### How was it tested?

Unit tests added in `k8s/dynamic_test.go` (per-group routing, subresource filter, partial-error handling, `ForceRefresh`). End-to-end verified by running the example operator (https://github.com/grafana/app-sdk-example-todo-app) against a local k3d cluster with managed kinds routed to different apiservers — the error above no longer reproduces.

Godoc on the new/changed types and constructors.

### Notes to Reviewers

Breaking changes:

- `resource.ClientGenerator` gains `DiscoveryClient() (DiscoveryClient, error)` — out-of-tree implementations must add it.
- `NewDynamicPatcher(*rest.Config, interval)` → `NewDynamicPatcher(ClientGenerator, interval)`.
- `DynamicPatcher.Get`/`Patch` + `DynamicKindPatcher.Get`/`Patch` return `resource.Object` (concrete type `*resource.UntypedObject`) instead of `*resource.UnstructuredWrapper`.
- `DynamicPatcher.ForceRefresh()` only refreshes groups already queried; no-op on a fresh patcher.

---
PR description was generated by Claude Code.

### DIFF
--- a/resource/client.go
+++ b/resource/client.go
@@ -6,8 +6,8 @@ import (
 	"net/http"
 	"net/url"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/rest"
 )
 
 const NamespaceAll = ""
@@ -281,9 +281,11 @@ type ClientGenerator interface {
 	// GetCustomRouteClient returns a Client for the provided GroupVersion. This returned Client is not guaranteed to be unique,
 	// and can be shared by other ClientForGV calls.
 	GetCustomRouteClient(schema.GroupVersion, string) (CustomRouteClient, error)
-	// KubeConfigForGroup returns the rest.Config to use for clients that operate on the given API group.
-	// Implementations should apply any per-group routing configuration (such as a KubeConfigProvider),
-	// so callers that only know the group (e.g. DynamicPatcher) can still reach the correct API server
-	// in multi-host setups.
-	KubeConfigForGroup(group string) rest.Config
+}
+
+// DiscoveryClient inspects the API groups and resources exposed by the underlying storage system.
+type DiscoveryClient interface {
+	// APIGroupInfo returns the preferred-version resources exposed by the given API group.
+	// Each entry's Group and Version fields are populated with the resource's preferred GroupVersion.
+	APIGroupInfo(apiGroup string) ([]metav1.APIResource, error)
 }
--- a/resource/client.go
+++ b/resource/client.go
@@ -287,9 +287,10 @@ type ClientGenerator interface {
 	DiscoveryClient() (DiscoveryClient, error)
 }
 
-// DiscoveryClient inspects the API groups and resources exposed by the underlying storage system.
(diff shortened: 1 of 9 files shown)


resource/client.go
```diff
@@ -280,4 +281,15 @@ type ClientGenerator interface {
...
+	// underlying storage system. Implementations may keep per-group clients internally to handle
+	// setups where different groups are routed to different hosts.
+	DiscoveryClient() (DiscoveryClient, error)
+}
+
+// DiscoveryClient inspects the API groups and resources exposed by the underlying storage system.
+type DiscoveryClient interface {
+	// APIGroupInfo returns the preferred-version resources exposed by the given API group.
+	// Each entry's Group and Version fields are populated with the resource's preferred GroupVersion.
+	APIGroupInfo(apiGroup string) ([]metav1.APIResource, error)
```
IfSentient: If we're returning only the preferred versions, can we name this something to better reflect that behavior, like `PreferredVersionResources`? 

We could also just return a single [metav1.APIResourceList](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#APIResourceList) so that it is more clear that the `GroupVersion` is the same for all kinds in the response (as the `preferredVersion` is set at an APIGroup level).

IfSentient (commented):
Looks great, just one comment on naming

resource/client.go
amalavet: https://github.com/grafana/grafana-app-sdk/commit/0b78d25b80452dc7880dfd83589392b852d0eb3b

IfSentient (approved):

PR #1361, opened 2026-04-24
Chore: require golang 1.26

ryantxu:
K8s 1.36 requires golang 1.26 -- rather than upgrade both in one PR, this splits the go-upgrade (and relevant lint fixes) into one PR without changing k8s


resource/simplestore.go
```diff
@@ -52,6 +52,7 @@ func WithResourceVersion(resourceVersion string) ObjectMetadataOption {
 // allowing the user to work with the actual type in the Schema Object's spec,
 // without casting in and out of the Object interface.
 // It should be instantiated with NewSimpleStore.
+//
```
ryantxu, line 55: lint wants Deprecated to be its own paragraph

codegen/jennies/manifest.go
```diff
@@ -10,28 +10,32 @@ import (
...
 )
 
 const (
 	VersionV1Alpha1 = "v1alpha1"
 	VersionV1Alpha2 = "v1alpha2"
 )
 
+const (
+	keyMetadata = "metadata"
+	keySpec     = "spec"
```
ryantxu: lint did not like how often the string constants where used in this file

.gitignore
```diff
@@ -14,6 +14,9 @@
 # Dependency directories (remove the comment below to include it)
 vendor/
 
+# VS Code settings
+.vscode
```
ryantxu, line 18: unrelated, but helps me :)

operator/opinionatedwatcher.go
```diff
@@ -4,10 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/codes"
-	"k8s.io/utils/strings/slices"
```
ryantxu, line 10: now deprecated!

operator/informer_concurrent.go
```diff
@@ -115,7 +115,7 @@ func (ci *ConcurrentInformer) Run(ctx context.Context) error {
 }
 
 func (ci *ConcurrentInformer) HealthChecks() []health.Check {
-	checks := make([]health.Check, 0)
+	checks := make([]health.Check, 0, 2)
```
IfSentient, line 118: Since we don't know the size in advance of the health checks from the informer, we may want to just do a `nolint:prealloc` here. Though setting the capacity to 2 doesn't hurt anything because `append` will grow the capacity if required.

app/runner.go
```diff
@@ -123,7 +123,7 @@ func (m *MultiRunner) PrometheusCollectors() []prometheus.Collector {
 
 // HealthChecks implements health.Checker
 func (m *MultiRunner) HealthChecks() []health.Check {
-	checks := make([]health.Check, 0)
+	checks := make([]health.Check, 0, len(m.Runners)*2)
```
IfSentient, line 126: Same here, we don't actually know the size of the runner's health checks, but the capacity being smaller shouldn't hurt anything.

IfSentient (approved):
LGTM, a few nits on the linter's capacity enforcement that are probably ok to ignore.

PR #1373, opened 2026-04-30
fix(codegen): selectable fields through CUE disjunctions

moustafab:
## What Changed? Why?

Selectable fields whose path crossed a CUE disjunction (e.g. `.spec.notificationSettings.receiver` where `notificationSettings` is `#SimplifiedRouting | #NamedRoutingTree`) generated Go code that didn't compile against the cog-generated union shape. cog represents each disjunction as a struct with one optional pointer field per variant, but the schema generator was emitting flat-path access.

This branch fixes that across a few commits:

- **Disjunction lookup** (`1e7b41f`): walk the variants of a CUE disjunction when a field doesn't resolve via plain LookupPath, so we can detect whether a selectable field exists at all.
- **String type casting** (`5a8152e`): always emit `string(...)` for string-typed selectables so named wrapper types (`type Foo string`) compile.
- **Selectable fields tests** (`5cb197d`): table-driven coverage for the existing path-based emission and the disjunction lookup.
- **Variant-aware access** (`cd466ff`): replace the discriminator-check approach with per-variant pointer checks matching cog's struct layout. Concrete-string variant fields (discriminators) are inlined as constants at codegen time. Three cases are handled:
  - discriminator field common to all variants
  - field present in only one variant
  - field present in some variants

For named disjunctions the variant Go field name comes from the CUE definition name (`#Foo` → `Foo`); for anonymous ones it follows cog's `UpperCamelCase(discField)+UpperCamelCase(discValue)` rule (`{type:"A",...}` → `TypeA`). The post-Decode pipeline wraps the field value in an AndOp from manifest constraints, so we recover variants by unwrapping AndOp operands and following ReferencePath into the named definition.

Motivating use cases:

- Discriminated union access — alerting/rules `AlertRule` selectors pass through the `#NotificationSettings` disjunction: [`apps/alerting/rules/kinds/alertRule.cue#L28-L37`](https://github.com/grafana/grafana/blob/main/apps/alerting/rules/kinds/alertRule.cue#L28-L37).
- Named wrapper types — alerting/rules `RecordingRule` selects `spec.metric` (`#MetricName`) and `spec.targetDatasourceUID` (`#DatasourceUID`): [`apps/alerting/rules/kinds/recordingRule.cue#L28-L34`](https://github.com/grafana/grafana/blob/main/apps/alerting/rules/kinds/recordingRule.cue#L28-L34).

### How was it tested?

- `go test ./codegen/...` — `TestGetSelectableFields` exercises all three cases for both inline and named disjunctions; `TestSchemaTemplateSelectableFieldCast` pins the template output for each case shape.
- `codegen/cuekind/testing/integration.cue` gains a `Notification` kind that exercises the named-disjunction shape end-to-end through the parser pipeline.
- Verified manually by regenerating the alerting/rules schemas (the motivating use cases above) and confirming the generated `*_schema_gen.go` compiles against the cog-generated spec types.

In-code comments on the new exported types (`UnionFieldAccess`, `UnionVariantAccess`) and on the helpers that needed a "why" (`disjunctionVariants` AndOp unwrap, `definitionName` ReferencePath quirk, `isFieldOptional` parent-lookup workaround).

### Notes to Reviewers

The bulk of the diff is in `codegen/jennies/schema.go` (new variant-walking helpers) and `codegen/templates/schema.tmpl` (per-variant emission branch). The discriminator-check shape from `1e7b41f` is fully replaced — `lookupInDisjunction` and its strategy 1/2/3 helpers are gone in favour of a single `disjunctionVariants` that walks AndOp/ReferencePath.

The variant Go field naming relies on cog's `DisjunctionOfAnonymousStructsToExplicit` pass for anonymous variants; if cog ever changes that rule we'll need to update `variantGoName` to match.


codegen/jennies/schema.go
```diff
@@ -93,37 +94,241 @@ func (*SchemaGenerator) getSelectableFields(kind *codegen.VersionedKind) ([]temp
...
+	}
+	return ""
+}
+
+func defNameFromSelector(sel cue.Selector) string {
+	s := sel.String()
+	if strings.HasPrefix(s, "#") {
+		return s[1:]
+	}
+	return ""
```
spinillos: Shouldn't be `return s`? You also can do something like [this](https://github.com/grafana/cog/blob/main/internal/simplecue/utils.go#L46).

IfSentient (dismissed):
LGTM with just the one change from @spinillos

codegen/jennies/schema.go
moustafab: Oh actually I think this function is somewhat unnecessary. The caller is relying on this behavior but I think it probably should just be inlined to avoid the misleading name.

IfSentient (approved):

PR #1389, opened 2026-05-14
fix(codegen): selectable fields through CUE disjunctions

harisrozajac:
## What Changed? Why?

This PR fixes code generation for `selectableFields` paths that cross a CUE union parent (disjunction), such as `spec.spec.name` on dashboard `VariableKind`. 

Without this fix, codegen fails validation for valid schemas or generates invalid Go field accessors for union variants.

I had [Global teamplate variables service](https://github.com/grafana/grafana/pull/122496) PR open for a couple of weeks, and noticed issues today when I merged main and tried generating a manifest. This behavior appears to have been introduced in https://github.com/grafana/grafana-app-sdk/pull/1373.

Global variables rely on server-side field selectors (for example `fieldSelector=spec.spec.name=status`) to resolve variables by logical name.

The dashboard variable spec is a CUE disjunction, but all variable kinds have `spec.name` as mandatory. In this shape:

- parent path lookup can fail before union fallback logic runs
- generated union variant accessor names can be wrong for anonymous disjunction variants

This blocks declarative `selectableFields` in CUE and breaks downstream generation/compilation for consumers like Grafana.

### How was it tested?

Tested in https://github.com/grafana/grafana/pull/122496 and verified that manifest was generated correctly with selectableFields.

Left some comments in the code. 

### Notes to Reviewers

I'm hoping we can support this, for now I have this [hack](https://github.com/grafana/grafana/pull/122496/changes#diff-1f1f4b228475141e37ea7e7af7ae7cf7e1bab4f809a85e4f43c7068267f5c5fd).


IfSentient (commented):
Can we add an example of this behavior to one of the golden test cue files in `codegen/cuekind/testing` (perhaps in [testkind.cue](https://github.com/grafana/grafana-app-sdk/blob/main/codegen/cuekind/testing/testkind.cue#L105)) and regenerate the codegen tests (`make regenerate-codegen-test-files`), just to have a committed example of the full output that we can note if it changes in a regression?

harisrozajac:
> Can we add an example of this behavior to one of the golden test cue files in `codegen/cuekind/testing` (perhaps in [testkind.cue](https://github.com/grafana/grafana-app-sdk/blob/main/codegen/cuekind/testing/testkind.cue#L105))

@IfSentient thank you for that suggestion 🙌 i've added an example 🙂

codegen/cuekind/testing/testkind.cue
```diff
@@ -175,5 +181,21 @@ testKind: {
...
+			schema: {
+				#UnionVariantA: {
+					kind: "VariantA"
+					spec: {name: string}
+				}
+				#UnionVariantB: {
+					kind: "VariantB"
+					spec: {name: string}
+				}
+				spec: #UnionVariantA | #UnionVariantB
```
IfSentient: It looks like this kind of union isn't generating a type at all via the go codegen: `codegen/testing/golden_generated/go/groupbygroup/testapp/v4/testkind_spec_gen.go` is empty. You may need to have the union be within the spec itself, rather than the root of the spec (such as `spec: { union: #UnionVariantA | #UnionVariantB }` and `spec.union.spec.name` for the selectableField).

IfSentient: This has probably revealed a different class of bug with the codegen where the root CUE value being a disjunction (the codegen roots each codegen step in the `spec`/`status`/etc. part of the schema) doesn't generate a type at all.

harisrozajac: good catch, and yeah i agree that that behavior looks like a codegen gap when the rooted spec value is itself a union. i reshaped the golden to use a nested union under spec so we still lock in union + selectable-field behavior with real generated types.

moustafab (approved):
LGTM 👍  Nice fix!

Was the difference after #1373 that this case would fail silently previously and now it errors? I'm not sure how the selectable fields could've been working for this case prior to that change, but afaict before it wouldn't have been blocking. Is that right or am I missing something?

harisrozajac:
@moustafab thanks for the review! 

> I'm not sure how the selectable fields could've been working for this case prior to that change, but afaict before it wouldn't have been blocking. Is that right or am I missing something?

Yeah selectable fields were working and CI was green and this was because prior to #1373 if the parent `LookupPath` failed (which it does for `spec.spec.name` on VariableKind), schema codegen silently skipped the field with no error, just no `WithSelectableFields` in *_schema_gen.go. Manifest generation is a separate path and still copied `selectableFields` straight from CUE, so make generate succeeded and the manifest got `SelectableFields: ["spec.spec.name"]`. API filtering also worked via hand-written `VariableToSelectableFields`.

Post-#1373: that silent skip became a hard error. The change is visible in:

https://github.com/grafana/grafana-app-sdk/blob/19ddf351c046672dd9a07957ed6df7bd312eabc5/codegen/jennies/schema.go#L97-L100 

@IfSentient do you mind approving this too? Looks like i can't merge without an approval from someone from grafana/grafana-app-platform-squad 🙂

IfSentient (approved):

PR #1415, opened 2026-06-11
feat: Serve App custom routes from operator.Runner

stevesg:
Adds support for serving an app.App's custom routes (non-CRUD endpoints)
through the same WebhookServer that already handles admission and conversion
webhooks reusing its HTTPS port and TLS config.


IfSentient (approved):
LGTM--in the future we may want to scope to just the routes declared instead of generic handlers, but I think this is a good place to start

PR #1431, opened 2026-06-30
Add searchFields to app manifest schema and codegen

pstibrany:
Adds a `searchFields` list to the per-kind manifest schema, parallel to the existing `selectableFields` and `additionalPrinterColumns`. Each entry declares a field exposed for search indexing and querying: a name, an optional JSON path into the resource, a value type (string, int64, double, boolean, date), the capabilities the field supports (filter, text, partial, sort, facet, retrieve, unranked), and optional flags for arrays, indexing the zero value when the path is absent, and a description.

The field is declared in the v1alpha2 CUE manifest schema and the codegen kind definition, generated onto `app.ManifestVersionKind`, and carried through the codegen pipeline into the embedded Go manifest. The hand-written v1alpha2 spec conversion maps it in both directions. This follows the same three-layer pattern (`codegen.SearchField`, `app.ManifestVersionKindSearchField`, generated `v1alpha2.AppManifestSearchField`) already used for additional printer columns.

No validation is performed yet beyond what the CUE type system enforces (allowed type and capability values are constrained by the schema). Path validation against the kind's OpenAPI schema and the capability matrix (text/partial/facet/sort require a string-typed field) are intentionally left out of this change and will follow separately.

`searchFields` is added to v1alpha2 only; v1alpha1 is left unchanged.

Part of grafana/search-and-storage-team#827.


app/appmanifest/v1alpha2/conversion.go
```diff
@@ -346,6 +373,33 @@ func SpecFromManifestData(data app.ManifestData) (*AppManifestSpec, error) {
...
+			if len(kind.SearchFields) > 0 {
+				k.SearchFields = make([]AppManifestSearchField, len(kind.SearchFields))
+				for i := 0; i < len(kind.SearchFields); i++ {
+					sf := kind.SearchFields[i]
+					translated := AppManifestSearchField{
+						Name:         sf.Name,
+						Type:         AppManifestSearchFieldType(sf.Type),
+						Capabilities: make([]AppManifestSearchFieldCapabilities, len(sf.Capabilities)),
+					}
+					for ci := 0; ci < len(sf.Capabilities); ci++ {
```
konsalex: nit: most places use `in .. range`

app/appmanifest/v1alpha2/appmanifest_spec_gen.go
```diff
@@ -419,6 +457,40 @@ func (AppManifestManifestVersionKindScope) OpenAPIModelName() string {
 	return "com.github.grafana.grafana-app-sdk.app.appmanifest.v1alpha2.AppManifestManifestVersionKindScope"
 }
 
+// +k8s:openapi-gen=true
+type AppManifestSearchFieldType string
+
+const (
+	AppManifestSearchFieldTypeString  AppManifestSearchFieldType = "string"
```
konsalex, line 469: nit: Should we add a link to Grafana repo (if any), where those types are defined? So it will be easier to keep them in-sync if we need to update with a new type?

konsalex:
Just nits, leaving @IfSentient to review

app/appmanifest/v1alpha2/conversion.go
pstibrany: Done — switched both loops in the v1alpha2 conversion to `range`.

app/appmanifest/v1alpha2/appmanifest_spec_gen.go
pstibrany: That file is generated, so I added the reference on the `#SearchField` definition in the CUE source instead — it now renders above `AppManifestSearchField` here and in the OpenAPI, and survives regeneration. It points at Grafana's `SearchFieldType`/`SearchCapability` as the values to keep in sync.

IfSentient (approved):

PR #1439, opened 2026-07-14
feat(deployment): add kinds to support enterprise/cloud router

charandas:
## What Changed? Why?

Introduces RouteBackend in v1alpha2 which in addition to ManifestData will power the cloud routes loader.


app/appmanifest/v1alpha3/appmanifest_client_gen.go
ryantxu, line 1: IIUC, the v1alpha3 package can be removed (left over from previous. codegen)

ryantxu (approved):

PR #1449, opened 2026-07-20
chore(deps): consolidate renovate updates

amalavet:
## What
Consolidates open dependency update PRs:

- #1446 chore(deps): bump the all group across 1 directory with 14 updates
- #1407 ci(actions): update github actions
- #1445 chore(deps): bump the all group across 1 directory with 6 updates
- #1409 build(deps): update go dependencies (only k8s.io/gengo/v2 and unrelated deps were still stale; the rest were already superseded)
- #1408 build(k8s): update kubernetes dependencies (only k8s.io/gengo/v2 was still stale)
- #1405 chore(deps): bump github.com/grafana/codejen from 0.0.4-0.20250428224353-8764ced07770 to 0.0.4 in the cue group across 1 directory (already superseded, no-op)

Not included:
- #1447, #1441, #1437 targeted lts/v0.24, which we no longer maintain — closed instead of merging
- cuelang.org/go bump from #1409 (breaks github.com/grafana/cog, left at current version)

PR #1476, opened 2026-08-14
fix: keep enum and other constraints when flattening allOf in CRD schemas

pstibrany:
Generated CRDs silently lost `enum` constraints starting with v0.57.0. For example `AlertRule.spec.notificationSettings.type` in Grafana went from `enum: [SimplifiedRouting, NamedRoutingTree]` to `allOf: [{}]` with a `default`, so the API server no longer restricted the field to the two valid values.

CRD schemas have to be structural, so `resolveSchema` merges `allOf` entries into the parent schema. That merge only copied `type`, `properties`, `items` and `additionalProperties` — `enum`, `pattern`, `format`, `minLength`, minimum/maximum, `description` and friends were dropped on the floor.

Nothing noticed until now because these fields used to be emitted as a plain `$ref`, which is inlined as a whole. cog v0.1.23 (pulled in by #1449) changed that: a field which references another type and also has a default is now emitted as an `allOf` with a single `$ref`, since in OpenAPI 3.0 a `$ref` cannot have siblings.

The fix copies the remaining validation and documentation fields during the merge, but only where the target schema does not set them itself. It also stops adding `allOf`/`anyOf` entries that carry no required fields, which only produced an empty object in the CRD.

The kube-openapi path is unaffected — it keeps `allOf` and refs as they are.

Verified by regenerating `apps/alerting/rules` in Grafana: the enum is back, and the only difference from the v0.56.6 output is the added `default: SimplifiedRouting`, which is correct.


pstibrany:
[AI-generated] Reviewed the regression test and the regenerated golden files, looks good to me.

Confirmed the test guards the regression: keeping your commit but reverting only `app/manifest.go` makes `TestCRDGenerator/JSON`, `TestCRDGenerator/YAML` and `TestCRDGenerator` fail. Nice that it covers both the oneOf merge (`spec.union`) and a plain single-type reference (`spec.map`).

I also ran the generated CRDs through `structuralschema.NewStructural` + `ValidateStructural` from apiextensions-apiserver: `customkind` v0-0 and v1-0 pass, as do all three alerting CRDs regenerated from Grafana `main`. Worth checking because Kubernetes rejects `default` inside `allOf`/`anyOf`/`oneOf`/`not`, and here the default sits on the root-level property with only `required`/`not` in the junctors.

Two things worth a conscious decision, neither blocking:

1. The default on a union discriminator changes admission behaviour vs v0.56.6. Before, `type` was required with no default, so a union object without `type` was rejected. Now it is defaulted to whichever branch cog emits first (`Type1` here, `SimplifiedRouting` in alerting), so `{group, details}` is accepted, stamped as `Type1`, and the generated Go `UnmarshalJSON` then decodes it as `Type1` and drops `details`. For alerting that is the desired behaviour, so I would keep it, but we could skip `Default` for `oneOf`/`anyOf` merges if we would rather not have a silent branch default.

2. The discriminator is not pinned per branch: the CUE says `type: #UnionTypeDiscriminator & "Type1"`, but the CRD says `enum: [Type1, Type2]` for both branches. That is cog collapsing the narrowing to a `$ref`, and v0.56.6 did the same via a bare `$ref`, so it is not a regression and out of scope here.

amalavet (approved):

PR #1478, opened 2026-08-14
fix(apiserver): preserve status code on custom route errors

cbodonnell:
## What Changed? Why?

This PR updates the custom route handler to preserve status code, reason, and message when the error returned is of type [`APIStatus`](https://pkg.go.dev/k8s.io/apimachinery@v0.36.2/pkg/api/errors#APIStatus), falling back to the current behavior otherwise. This allows custom route implementations to more ergonomically produce the desired http responses simply by returning an `APIStatus` error.

For example, instead of manually writing an http response code, the custom route could do:
```go
return apierrors.NewBadRequest("invalid request")
```

### How was it tested?

Added unit tests to cover both cases where an `APIStatus` error is returned and when another error type is returned.


k8s/apiserver/installer.go
```diff
@@ -689,12 +689,8 @@ func (r *defaultInstaller) registerResourceRouteOperation(ws *restful.WebService
...
 		a, err := r.App()
 		if err != nil {
-			resp.WriteHeader(http.StatusInternalServerError)
-			_ = json.NewEncoder(resp).Encode(metav1.Status{
-				Status:  metav1.StatusFailure,
-				Code:    http.StatusInternalServerError,
-				Message: err.Error(),
-			})
+			writeCustomRouteError(resp, err)
+			return
```
cbodonnell, line 693: Are there any backwards compatibility concerns with changing this behavior that previously _always_ returned a 500 status? This could alternatively be an opt-in behavior when registering a custom route?

jliuhq (dismissed):

ryantxu (dismissed):

ryantxu (approved):

PR #1499, opened 2026-08-24
feat(manifest): Validate we do not conflict with /search or /trash

ryantxu:
## What Changed? Why?

grafana/grafana now adds built-in resource search using the path `POST <plural>/search`, this tells people loudly and early that registering the same path will conflict.


pstibrany (dismissed):
Looks good overall. 

One non-blocking note: `/search` and `/trash` remain reserved even when a kind opts out of the built-in endpoint. I think that is fine, but it may be worth documenting that explicitly in a test or comment.

stevesg (dismissed):

stevesg (approved):

PR #1516, opened 2026-08-27
feat(plugins): add plugin.Run helper for App SDK backends

stevesg:
Wraps plugin `app.Manage`, registering the appadapter plugin.v3 services
and running the App's Runner for the lifetime of the plugin.

	func main() {
		if err := plugin.Run(myapp.Provider()); err != nil {
			log.DefaultLogger.Error(err.Error())
			os.Exit(1)
		}
	}

Part of https://github.com/grafana/grafana-app-platform-squad/issues/111

### DIFF
--- a/plugin/run.go
+++ b/plugin/run.go
@@ -1,0 +1,173 @@
+package plugin
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	backendapp "github.com/grafana/grafana-plugin-sdk-go/backend/app"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/instancemgmt"
+	backendlog "github.com/grafana/grafana-plugin-sdk-go/backend/log"
+	"k8s.io/client-go/rest"
+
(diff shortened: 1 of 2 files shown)


plugin/run.go
```diff
@@ -0,0 +1,178 @@
...
+	kubeConfig rest.Config
+	pluginID   string
+	appFunc    backendapp.InstanceFactoryFunc
+	manageOpts backendapp.ManageOpts
+}
+
+type stubInstance struct{}
+
+func (stubInstance) CheckHealth(_ context.Context, _ *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
+	// TODO: Should we use the healthchecks from app.App?
```
stevesg, line 164: I've not dug into whether this is something we should do or not - open to opinions.

ryantxu: the CheckHealth, etc will be useful/required when integrating with an existing app-plugin.  

We will also need to figure out how to run an app-plugin **without** these implemented -- I think it would look broken now

ryantxu (approved):

PR #1519, opened 2026-08-28
Add an authz section to custom routes

IfSentient:
## What Changed? Why?

Add an authz section to custom routes which generates grafana authz extensions in the OpenAPI for the route in the AppManifest.

These extensions are not currently used by grafana, but will be in the future for apps.

### How was it tested?

Codegen tests

Not documented yet aside from in the CUE, as the functionality isn't yet implemented grafana-side


konsalex (approved):
LGTM

PR #1537, opened 2026-09-07
feat(plugins): Derive KubeConfig from service account env vars

stevesg:
Run() now falls back to a rest.Config built from GF_APP_URL and
GF_PLUGIN_APP_CLIENT_SECRET when WithKubeConfig is not used.

Part of https://github.com/grafana/grafana-app-platform-squad/issues/111

### DIFF
--- a/plugin/serviceaccount.go
+++ b/plugin/serviceaccount.go
@@ -1,0 +1,39 @@
+package plugin
+
+import (
+	"context"
+
+	"github.com/grafana/grafana-plugin-sdk-go/config"
+	"k8s.io/client-go/rest"
+)
+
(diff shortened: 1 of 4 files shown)


plugin/serviceaccount.go
```diff
@@ -0,0 +1,39 @@
+package plugin
+
+import (
+	"context"
+
+	"github.com/grafana/grafana-plugin-sdk-go/config"
+	"k8s.io/client-go/rest"
+)
+
+// serviceAccountKubeConfig builds a rest.Config from the plugin's provisioned
```
ryantxu, line 10: 🤔  IIUC, service accounts are tied to a single stack/org/namespace, so it does not "work" in multi-tenancy

PR #1541, opened 2026-09-08
feat(codegen): let kinds declare embeddable fields and hybrid search

owensmallwood:
**What this does**
This will allow you to create embeddings for unified storage resources and expose them through a `/search/hybrid` endpoint. You define which fields are embeddable in the cue.

**Example cue**
```cue
package example

manifest: {
	appName: "example-app"

	embed: {
		// One manual re-embedding version shared across all API versions of articles.
		// Increase it to request a backfill of existing resources.
		// Changing declared inputs alone does not require a version bump.
		//
		// Custom embedding builders define their inputs and version in Go.
		// They omit this entry and the per-version embed block below.
		articles: {reembedVersion: 1}
	}

	versions: v1: kinds: [{
		kind: "Article"
		schema: spec: {
			title:    string
			body:     string
			category: string
		}

		// Each API version defines its own extraction paths.
		// These fields are used only to generate the text to be embedded.
		// Declaring them does not enable filtering embeddings by these fields.
		embed: fields: [
			{name: "title", path: "spec.title"},
			{name: "body", path: "spec.body"},
		]

		// Search fields are independent of embedding inputs:
		// category supports filtering without being embedded.
		searchFields: [{
			name:         "category"
			path:         "spec.category"
			type:         "string"
			capabilities: ["filter"]
		}]

		// Declares the hybrid search endpoint. Defaults to false.
		search: hybrid: true
	}]
}
```

**Important notes**
- Creating embeddings follows a similar pattern to our search document builders. Complex resources like dashboards have custom builders in Go code. These all have a `version` integer in them. Increasing this version signals that we need to re-embed all of that resource. This PR follows the same approach.
- There will be more follow-up PRs. I'm starting with the app-sdk changes first since it needs to be agreed on first. The rest is just implementation after this.


codegen/cuekind/def.cue
```diff
@@ -224,6 +231,28 @@ SchemaWithOperatorState: Schema & {
...
+	// change to those fields should re-embed resources that already exist:
+	// flagging or unflagging a field, changing a path, reordering them.
+	//
+	// Bumping re-embeds every resource of this kind on every instance, which the
+	// search backend pays for in provider spend and backfill time, so leave it
+	// alone for changes not worth paying for.
+	//
+	// One number per kind: declare the same value on every version. Carry the
+	// current value over when adding a version, rather than restarting at 1.
+	version: int & >0
```
owensmallwood: I think this is probably the most controversial part of this PR. The problem its solving is that we need a way to trigger re-embedding for that resource when the embedding schema/format changes. For example, adding a new field that we want embedded. 

The awkward part is the embed version has to be declared for the kind, across api versions (I think it could easily be declared as a variable across all api versions to keep things simple).

I considered doing this automatically when a schema change is detected, but I think triggering this should be intentional by a human since it could be expensive. There are cases where a re-embed is not necessary, like reordering fields or adding/removing an embedded field that is minor.

The plan is to guard this with a CI step. So if the embed version gets bumped to trigger a re-embed, the CI step will fail unless a certain label is present.

pstibrany:
I don’t think embedding configuration should reuse `searchFields`. Search fields and embedding inputs are independent: not every search field should be embedded, and a field may need to be embedded without being exposed as a search field. Pathless fields supplied by custom builders are another case that the current validation excludes.

I’d prefer a separate `vectorSearch` or `embedding` section that defines the inputs used to construct the embedding document.

I also think `embed.version` is redundant for declarative fields. We can calculate a stable hash from the ordered embedding-field configuration and use changes to that hash to trigger re-embedding. Custom Go builders may still need their own explicit version because their behavior can change independently of the manifest.

`search.hybrid` looks fine to me.

codegen/jennies/embed_fields.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/grafana/grafana-app-sdk/codegen"
+)
+
+func validateEmbedFields(vk codegen.VersionedKind, version string) error {
+	if vk.Embed == nil || !vk.Schema.Exists() {
+		return nil
+	}
+	for _, field := range vk.Embed.Fields {
+		// A custom builder supplies pathless fields; they need no resource schema entry.
```
owensmallwood: if theres a custom builder we dont need to define embeddable fields anyways

owensmallwood:
@codex

owensmallwood:
@pstibrany I've kept the content version field in the cue, but made it global to the app. I don't want to automatically re-embed when the embed fields change, since this doesn't always mean embedding is needed. It should be something triggered intentionally by a human as it can be an expensive operation and should be used sparingly.

pstibrany:
Thanks, this looks much better. Separating embedding inputs from `searchFields` addresses my main concern, and omitting this section for custom builders makes sense.

As discussed yesterday, an explicit version for manual re-embedding makes sense. Would `reembedVersion` be clearer than `contentVersion`? `contentVersion` sounds like it identifies the exact input definition and therefore should change whenever the fields change. `reembedVersion` more directly describes its purpose: intentionally triggering re-embedding of existing resources.

pstibrany:
The opening description still says that users define which `searchFields` are embeddable. That is no longer accurate now that embedding inputs have their own independent section. Could we update it to match the new design?

codegen/jennies/embed_fields.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			return fmt.Errorf("kind %q version %q embed field %q: %w", vk.Kind, version, field.Name, err)
+		}
+	}
+	return nil
+}
+
+func embedFieldSupportsText(leaf cue.Value) bool {
+	kind := leaf.IncompleteKind() &^ cue.NullKind
+	if kind == 0 || kind&cue.StringKind != 0 {
```
pstibrany: Could this accept a mixed type such as `string | int`? `IncompleteKind()` would contain both bits, so checking `kind&cue.StringKind != 0` appears to accept it. The same applies to `embedScalarSupportsText`.

Since the declared contract permits only strings and string arrays, should this require the non-null kind to equal `cue.StringKind`? It would also be useful to add test cases for `string | int` and `[...(string | int)]`.

app/appmanifest/v1alpha2/conversion.go
```diff
@@ -412,6 +435,17 @@ func SpecFromManifestData(data app.ManifestData) (*AppManifestSpec, error) {
...
+			if kind.Embed != nil {
+				k.Embed = &AppManifestManifestVersionKindEmbed{}
+				if kind.Embed.Fields != nil {
+					k.Embed.Fields = make([]AppManifestManifestVersionKindEmbedField, len(kind.Embed.Fields))
+					for i, field := range kind.Embed.Fields {
+						k.Embed.Fields[i].Name = field.Name
+						k.Embed.Fields[i].Path = field.Path
+					}
 				}
 			}
```
RafaelPaulovic, line 450: [P2] Emit an empty array for zero-value embed declarations: When callers provide the valid zero-value declaration `kind.Embed = &app.ManifestVersionKindEmbed{}`, `ManifestData.Validate` accepts it, but this conversion leaves `Fields` nil. JSON serialization then emits `"fields": null`, while the generated `v1alpha2` schema requires fields to be an array, causing the API server to reject the manifest. Initialize `Fields` to an empty slice whenever `Embed` is non-nil.

pstibrany:
The opening has a duplicated “which.” The example also still uses `contentVersion` and calls it a content version. Could we update both to `reembedVersion` and describe it as the manual re-embedding version?

codegen/jennies/embed_fields.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	for _, field := range vk.Embed.Fields {
+		if field.Path == "" {
+			return fmt.Errorf("kind %q version %q embed field %q requires a path", vk.Kind, version, field.Name)
+		}
+		if !vk.Schema.Exists() {
+			continue
+		}
+		leaf, resolved, err := resolveSearchFieldPath(vk.Schema, field.Path)
+		if err == nil && resolved && !embedFieldSupportsText(leaf) {
```
pstibrany: When a path exists in only one branch of a CUE disjunction, `resolveSearchFieldPath` returns `resolved=false`. This skips the type check here, so a path pointing to an integer in that branch could be accepted as an embedding input.

Should embedding validation inspect the matching branch or reject paths whose type cannot be resolved?

codegen/jennies/embed_fields.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			return fmt.Errorf("kind %q version %q embed field %q: %w", vk.Kind, version, field.Name, err)
+		}
+	}
+	return nil
+}
+
+func embedFieldSupportsText(leaf cue.Value) bool {
+	kind := leaf.IncompleteKind() &^ cue.NullKind
+	if kind == 0 || kind == cue.StringKind {
```
pstibrany: Does `kind == 0` cause a null-only field to be accepted? After removing `cue.NullKind`, a `null` value appears to produce zero here. The same may apply to null-only list elements.

Since these fields cannot contribute text, should they be rejected? Tests for `null` and `[...null]` would help clarify the intended behavior.

RafaelPaulovic (dismissed):
LGTM, thank you

RafaelPaulovic (approved):

IfSentient (approved):

pstibrany (approved):
Thanks for addressing all of our feedback!

Issue #1559, opened 2026-09-17
`const` is dropped when building schemas for kubernetes, so `oneOf` branches lose what tells them apart


gillesdemey:
## What happens

When a kind's schema has a `oneOf`, the `oneOf` reaches the served OpenAPI fine. But the
`const` on each branch — the fixed value that tells one branch from another — is thrown
away. Nothing warns about it, and the generated Go and CRD schema both look fine.

Clients end up with a union whose branches they can't tell apart.

## Reproducer

This CUE:

```cue
#EmailV1Integration: {type: "email", version: "v1", settings: {addresses: string}}
#SlackV1Integration: {type: "slack", version: "v1", settings: {recipient?: string}}

#Integration: #EmailV1Integration | #SlackV1Integration
```

generates the manifest schema correctly:

```json
"type":    {"const": "email", "type": "string"},
"version": {"const": "v1",    "type": "string"}
```

Then `VersionSchema.AsKubeOpenAPI(...)`, which is what the apiserver uses to serve
`/openapi/v3`, gives back:

```json
"type":    {"type": "string"},
"version": {"type": "string"}
```

The `oneOf` is still there. The fixed values are gone. A one-value `enum` written by
hand does survive, which is what makes this look accidental.

## Why

`spec.SchemaProps` in `k8s.io/kube-openapi` has no `Const` field — `const` arrived with
OpenAPI 3.1, after those types. Two functions build schemas, both copy `Enum`, and
neither has anywhere to put `Const`:

- `oapi3SchemaToKubeSchema` (`app/manifest.go:1108`) — serves `/openapi/v3`
- `resolveSchema` (`app/manifest.go:1333`) — builds the CRD schema

The value is there to read (`openapi3.Schema.Const`). It just isn't read.

## Why it matters

In the generated TypeScript client it's the difference between

```ts
type Integration = EmailV1Integration | SlackV1Integration;  // can narrow on .type
```

and a union you can't narrow, because every branch is just `{ type: string, ... }`.
Grafana's contact points hand-write that union in TypeScript to get around it.

Easy to confuse with this but not the same: `rules.alerting`'s
`AlertRuleNotificationSettings` also comes out without a fixed value on `type`. That one
writes `#NotificationSettingsType & "SimplifiedRouting"`, which codegen turns into
`default` plus an `allOf`, so no `const` is produced at all. Fixing `const` won't help
it unless the value is written inline.

## Suggested fix

[...]

PR #1560, opened 2026-09-17
fix(app): keep const when building schemas for kubernetes

gillesdemey:
## What Changed? Why?

Addresses Issue #1559

When a kind's schema has a `oneOf`, the `const` on each branch is dropped on the way to
the served OpenAPI. The `oneOf` survives, the fixed values don't — so clients get a
union whose branches they can't tell apart. In the generated TypeScript that's the
difference between a union you can narrow on `.type` and one where every branch is just
`{ type: string, ... }`.

`spec.SchemaProps` in `k8s.io/kube-openapi` has no `Const` field, since `const` arrived
with OpenAPI 3.1, after those types. Both functions that build schemas copy `Enum` and
have nowhere to put `Const`, so the value disappears with no warning:

- `oapi3SchemaToKubeSchema` — serves `/openapi/v3`
- `resolveSchema` — builds the CRD schema

This writes it out as a one-value `enum` instead. Same meaning, and `enum` exists in
both targets — CRD schemas allow `enum` but not `const`. An `enum` already on the schema
wins, so this only fills a gap.

`const` is left alone in the manifest itself, which is a 3.1 document where it's the
better spelling.

### How was it tested?

Four new table cases, two per function. All four fail before this change and pass after.

- `TestVersionSchema_AsKubeOpenAPI` — `const` becomes `enum: ["bar"]`; an existing `enum`
  wins over `const`
- `TestGetCRDOpenAPISchema` — the same two through the CRD path

I also checked the case that prompted this, a receiver-integration union from CUE.
Before, `AsKubeOpenAPI` gave `"type":{"type":"string"}`; after,
`"type":{"type":"string","enum":["email"]}` with the `oneOf` intact — enough for
`@rtk-query/codegen-openapi` to emit `type: 'email'`.

`go test ./app/...` and `golangci-lint run ./app/...` (v2.13.2) are clean.

A comment at each of the two spots, on why the value is rewritten rather than copied. No
docs change — `oneOf` in a kind schema already works, this just stops it losing
information on the way out.

### Notes to Reviewers

The second hunk is the one worth a look. `resolveSchema` builds the CRD schema, and I've
assumed rewriting to `enum` is better than dropping the constraint. If you'd rather that
path stayed as it is, the first hunk on its own fixes the served-OpenAPI problem.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gillesdemey:
Blocker for grafana/grafana#133106 — that PR relies on a `oneOf` discriminator reaching the generated TypeScript client as a literal type. Without this fix the `const` is dropped when the manifest schema is converted for kubernetes, so the served OpenAPI has `variant: {"type": "string"}` and the TypeScript union cannot be narrowed, which is the whole point of the change.

PR #1565, opened 2026-09-22
Search: add string map field type

RafaelPaulovic:
Search manifests cannot currently declare map-shaped values, which forces applications to flatten maps in custom document builders and hides keyed filtering from the field model.

Tracks https://github.com/grafana/search-and-storage-team/issues/930.

This adds the `stringMap` search field type for `map[string]string` values. Code generation validates the schema shape, rejects `array: true`, and limits capabilities to `filter` and `retrieve`. Generated app-manifest schemas and focused validation tests are included.

Validation:

- `make lint`
- `go test ./searchfields ./codegen/jennies ./app/appmanifest/v1alpha2`


IfSentient (approved):

pstibrany:
For a regular struct, `LookupPath(cue.AnyString)` does not exist, so this returns `nil`; `searchFieldTypeCompatibility` then skips validation. That means a field such as `{name: string, count: int64}` can be declared as `stringMap` and pass codegen even though it is not a `map[string]string`. Could we return an incompatible result for unrecognized structs when the declared type is `stringMap`, and add a mixed-struct test?

RafaelPaulovic:
We discussed internally and choose deferring a first-class `stringMap` field type until another kind needs it. If another team needs the same we can revisit this PR.

What this means for alerting team: they can continue using its existing custom search builder and flattened key / `key=value` representation.

